PR #2, opened 2018-09-12
[#1] Add support for processing ECC certificates as part of the trust chain

### LINKED ISSUE #1: ACA Import Trust Chain CA Certificates needs to support ECC based certificates.
Even though the current version of the ACA only supports RSA public key based EK certs, TPM manufacturers can use Ecc keys to sign RSA public keys within an EK certs. Specifically Nuvoton_TPM_Root_CA_2110.cer is an ECC based certificate which is used to validate some Nuvoton produced RSA based EK certs. 
```
EK cert:
Subject Public Key Info:
            Public Key Algorithm: rsaEncryption
                Public-Key: (2048 bit)
Signature Algorithm: ecdsa-with-SHA256
```
The ACA should be able to upload an Ecc based intermediate or root Trust Chain CA Certificate. Furthermore the ACA should be able to validate and use the ECC based certs to validate the Certificate chain of any EK certificate.

apldev3:
This is a small merge request to allow for uploading of ECC certificate to the AttestationCA Portal.

Closes #1

### DIFF
--- a/HIRS_AttestationCA/src/main/java/hirs/attestationca/AbstractAttestationCertificateAuthority.java
+++ b/HIRS_AttestationCA/src/main/java/hirs/attestationca/AbstractAttestationCertificateAuthority.java
@@ -233,9 +233,13 @@ public abstract class AbstractAttestationCertificateAuthority
                     this.certificateManager, ecBytesFromIdentityRequest
             );
             try {
-                ekPublicKey = assemblePublicKey(
-                    Certificate.getPublicKeyModulus(
-                            endorsementCredential.getX509Certificate()).toByteArray());
+                BigInteger publicKeyModulus = Certificate.getPublicKeyModulus(
+                        endorsementCredential.getX509Certificate());
+                if (publicKeyModulus != null) {
+                    ekPublicKey = assemblePublicKey(publicKeyModulus.toByteArray());
+                } else {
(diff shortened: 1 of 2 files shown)


iadgovuser26:
Tested with  Nuvaton TPM Model NPCT6xx using the Nuvoton TPM Root CA 2110 with all policies enabled. Worked!

apldev4:
The `publicKeyModulusHexValue` field of `Certificate` is never used anywhere. I'd rather remove it than set it to null. Setting variables to null because there is not a clear use for them is asking for trouble. The fact that `Certificate.getPublicKeyModulus()` can now return a null won't cause a NullPointerException until someone passes an ECC Endorsement Credential as part of their identity request.

apldev3:
@apldev4 there's actually a somewhat indirect use for them if you poke around a bit. The field is used in `CertificateSelector` as part of the 1.2 provisioning process in the ACA.

apldev4:
> @apldev4 there's actually a somewhat indirect use for them if you poke around a bit. The field is used in `CertificateSelector` as part of the 1.2 provisioning process in the ACA. Trust me, I wanted to remove the field, but it appears to be a rather integral part of the original provisioning.

Good call. Looks good to me, then.

apldev3:
Got two thumbs up, merging in now.

Issue #7, opened 2018-09-17
Feature request: HIRS support Ubuntu 18.04 LTS


mwiseman32:
Both ACA and and Provisioner* support for Ubuntu 18.04 LTS using repo tpm tools.

iadgovuser26:
The [HIRS build pages](https://github.com/nsacyber/HIRS/wiki/Hirs-build-guide) imply that building with Debian (and hence Ubuntu) is supported. However the section on building the TPM 2.0 Provisioner is missing instructions on building in Debian. I attempted to [build the TPM 2.0 Provisioner](https://github.com/nsacyber/HIRS/blob/master/HIRS_ProvisionerTPM2/README.md) on Ubuntu 18.04.1 and it failed.

Building via gradle currently attempts to build all projects s the TPM2 provisioner build will need to work. The package script utilizes gradle build so no packages can be built until the TPM2 provisioner builds.

apldev3:
@iadgovuser26 testing the Gradle build locally after some small updates. If it works, I'll be pushing up a merge request to at least get the build to work. Packaging into a Debian format will still need additional time, but it should stop overall build from failing.

apldev3:
With the latest push, we have a first crack at supporting Ubuntu 18.04 Code-wise. The next step will be to test the code in a live environment and, pending its success, then work to make the DEB package.

iadgovuser26:
I some trouble  building the HIRS_ProvisionerTPM2 project with this commit. 
Error was with cmake, at build/lib/cpr-src/opt/CMakelists.txt:60.
Had to delete the build directory , then it seemed to work fine.

apldev3:
With the current Merge Request that's up, Ubuntu building/packaging has been added, and support for Ubuntu 17.10 is confirmed, but there seems to be an issue with the performing the last major step of the provisioning process on Ubuntu 18.04. Will update as more becomes known.

apldev3:
Before closing this ticket, I feel an update is pertinent. Current investigation towards supporting Ubuntu 18.X releases out-of-the-box is not promising. There seems to be a bug(s) relating to the version of [tpm2-tools](https://github.com/tpm2-software/tpm2-tools) that is natively available to Ubuntu 18.X. As of 18.04 the version was updated to 2.1.0.

This particular version is known to have some bugs in major tools, and it has been suggested that users upgrade.

[...]

PR #8, opened 2018-09-17
[#7] Support Building (Not Packaging) on Ubuntu 18.04

apldev3:
This is a small merge request that ensures that the full Gradle build will complete in an Ubuntu 18.04 development environment. In theory, it should also work on Ubuntu 17.10, but this has not been formally tested.

In order to test this MR, an Ubuntu development environment needs to be built with all of the dependencies that have been listed in the new `developer-dependencies-debian.md` file. Once they've been installed the usual build instructions should be followed for building the TPM 2.0 Provisioner and the rest of the HIRS project.

**NOTE:** This merge request does not provide support for TPM 2.0 provisioning in Debian environments. This is merely the first step in getting there. As such, this does not close issue the issue.

### DIFF
--- a/HIRS_ProvisionerTPM2/src/CommandTpm2.cpp
+++ b/HIRS_ProvisionerTPM2/src/CommandTpm2.cpp
@@ -411,7 +411,9 @@ string CommandTpm2::activateIdentity() {
         // Erase unnecessary zero padding due
         s.erase(2 + numBytesInCred, 134 - 2 - numBytesInCred);
         // Prepend header: MAGIC_NUMBER (0xBADCC0DE) + Version (0x00000001)
-        s.insert(s.begin(), {0xBA, 0xDC, 0xC0, 0xDE, 0x00, 0x00, 0x00, 0x01});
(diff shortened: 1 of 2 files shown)


HIRS_ProvisionerTPM2/src/CommandTpm2.cpp
```diff
@@ -411,7 +411,9 @@ string CommandTpm2::activateIdentity() {
         // Erase unnecessary zero padding due
         s.erase(2 + numBytesInCred, 134 - 2 - numBytesInCred);
         // Prepend header: MAGIC_NUMBER (0xBADCC0DE) + Version (0x00000001)
-        s.insert(s.begin(), {0xBA, 0xDC, 0xC0, 0xDE, 0x00, 0x00, 0x00, 0x01});
+        s.insert(s.begin(), {static_cast<char>(0xBA), static_cast<char>(0xDC),
+                             static_cast<char>(0xC0), static_cast<char>(0xDE),
```
apldev3: These static casts became necessary when building in the Ubuntu environment. There is a difference in the C++ compiler implementation that had gone unseen in the CentOS development environment, so this is required to fix a compilation issue.

PR #9, opened 2018-09-21
[#7] Setup Build/Package Support for Ubuntu

apldev3:
This merge request adds support for building and packaging on Ubuntu in general. The build guide will be updated corresponding to some of the new packaging scripts and the like as this gets merged in.

This is another incremental MR as it still doesn't quite close the issue. There is a known issue with the tool tpm2_activatecredential that's blocking the complete provisioning process that has to be investigated and resolved.

That being said, this MR does open the door for compiling and building on Ubuntu 17.10 as an alternative for the time being, which is using versions of TPM2-Tss and TPM2-Tools that are known to work with the provisioner.


apldev3:
As a quick note, the build has been tested on CentOS 7 to ensure no regressions occurred on that front.

apldev2:
lgtm

apldev3:
Rebased the branch to include testing on Travis CI.

HIRS_ProvisionerTPM2/src/Utils.cpp
```diff
@@ -232,25 +245,73 @@ namespace tpm2_tools_utils {
...
+    string distributionRelease;
+    ifstream releaseFile;
+    string line;
+    releaseFile.open("/etc/os-release");
+    if (releaseFile.is_open()) {
+        while (getline(releaseFile, line)) {
+            stringstream ss(line);
+            string item;
+            vector<string> tokens;
+            char* delim = const_cast<char*>("=");
```
apldev4: You can get rid of the variable `delim` and just do this:
```
while(getline(ss, item, '=')) {
```

HIRS_ProvisionerTPM2/src/CommandTpm2.cpp
```diff
@@ -133,7 +134,8 @@ void CommandTpm2::setAuthData() {
     stringstream argsStream;
 
     switch (version) {
-        case Tpm2ToolsVersion::VERSION_1_1_0: {
+        case Tpm2ToolsVersion::VERSION_1_1_0:
+        case Tpm2ToolsVersion::VERSION_2_1_0: {
```
apldev4: Curly braces are not needed around the lines of a single case -- only to enclose the entire switch statement. Flow control is managed by jumping to the correct case and then leaving the switch block when hitting the `break` statement for the case.

HIRS_ProvisionerTPM2/src/CommandTpm2.cpp
```diff
@@ -497,8 +499,12 @@ string CommandTpm2::createNvWriteCommandArgs(const string& nvIndex,
                           << " -a " << kDefaultOwnerAuthHandle
                           << " ";
 
-    if (version == Tpm2ToolsVersion::VERSION_1_1_0) {
-        argumentsStringStream << "-f ";
+    switch (version) {
+        case Tpm2ToolsVersion::VERSION_1_1_0:
+        case Tpm2ToolsVersion::VERSION_2_1_0: {
```
apldev4: These curly braces are not needed.

apldev3:
@apldev4 would you say that this is good to go at this point?

Issue #14, opened 2018-09-27
Optimize Travis CI Build


apldev3:
The Travis CI build is currently building/testing everything in about 13 minutes. This could be optimized if the subprojects are built/tested in parallel and a pre-built Docker image is provided to the test environments to prevent having to build a new Docker image on every test job.

apldev3:
#15 was merged and this issue can be closed.

PR #15, opened 2018-09-27
[#14] Parallelize Subproject CI Builds

apldev3:
This MR improves the Travis CI build time by taking advantage of their Build Matrix feature. This allows us to parallelize building/testing the subprojects all at the same time. In particular, this optimization shaves off about 4-5 minutes total.

Doesn't close the issue #14 just yet, as we're waiting to hear on if we can pre-build a Docker image and host it on Docker Hub.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,11 +5,22 @@ sudo: true
 
 language: java
 
+env:
+  matrix:
+    - SUBPROJECT=HIRS_Utils
+    - SUBPROJECT=HIRS_Provisioner
+    - SUBPROJECT=HIRS_ProvisionerTPM2
+    - SUBPROJECT=HIRS_Structs
+    - SUBPROJECT=HIRS_AttestationCA
+    - SUBPROJECT=HIRS_AttestationCAPortal
+    - SUBPROJECT=TPM_Utils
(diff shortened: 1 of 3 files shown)


apldev3:
Just pushed an update after having setup a Dockerhub repo. The Travis CI build now pulls a pre-built CentOS7 image that has all the dependencies already installed, such that they don't need to be installed everytime the CI job is run.

This should cut test run time by several additional minutes.

apldev3:
After the tests ran, we of course still have the issue with the pull request build since I added `tpm_module` as a subproject. The rest of the pull request build jobs all passed so that shouldn't be an issue.

Furthermore, having updated the build to pull from the pre-built Docker image on Docker Hub we shave off an additional minute or so. This seems to get the average build time down to roughly 7 minutes.

I would propose disabling the Build on Pull Requests feature on Travis CI since it doesn't seem to give much additional information and will occasionally fail like this, which raises unnecessary concerns and offers little to no additional information for the developers.

apldev3:
We've disabled the pull request builds so the build is now accurate in saying that the checks have all passed. Additionally, we can see the optimized gains in the following image. I'd say this ticket is good to be closed when this is merged.

image

Issue #6, opened 2018-09-17
ACA Should display the Platform Certificate Holder


iadgovuser26:
The Platform Certificate page does not currently display the holder information (EK certificate info). The holder is defined as:
```
 Holder ::= SEQUENCE {
     baseCertificateID   [0] IssuerSerial OPTIONAL,
            -- the issuer and serial number of
            -- the holder's Public Key Certificate
     entityName          [1] GeneralNames OPTIONAL,
            -- the name of the claimant or role
     objectDigestInfo    [2] ObjectDigestInfo OPTIONAL
            -- used to directly authenticate the
            -- holder, for example, an executable
 }

 IssuerSerial ::= SEQUENCE {
     issuer     GeneralNames,
     serial     CertificateSerialNumber,
     issuerUID  UniqueIdentifier OPTIONAL
 }
```
The baseCertificateID should be displayed with a hyperlink to the EK cert page (if it exists). Per the Platform Certificaate specification, "the BaseCertificateID choice MUST be used." As such, the entityName and objectDigestInfo optional entries, valid for attribute certificates in general, are never found in Platform Certificates.

apldev3:
#16 was merged. This can be closed at this point.

PR #16, opened 2018-09-28
[#6] Update Platform Credentials details page to display the Holder Information

cyrus-dev:
This merge request adds Holder information for platform credentials.  This will display the common name and associated holder serial number.  If the holder serial number is linked to an existing endorsement certificate, a link will be provided.  Otherwise just display the serial number of the endorsement link.

### DIFF
--- a/HIRS_Utils/src/main/java/hirs/data/persist/certificate/Certificate.java
+++ b/HIRS_Utils/src/main/java/hirs/data/persist/certificate/Certificate.java
@@ -196,6 +196,7 @@ public abstract class Certificate extends ArchivableEntity {
     public static final String HOLDER_SERIAL_NUMBER_FIELD = "holderSerialNumber";
     @Column(nullable = false, precision = MAX_NUMERIC_PRECISION, scale = 0)
     private final BigInteger holderSerialNumber;
+    private String holderIssuer;
 
     // we don't need to persist this, but we don't want to unpack this cert multiple times
     @Transient
@@ -220,6 +221,7 @@ public abstract class Certificate extends ArchivableEntity {
         this.certificateHash = 0;
         this.certAndTypeHash = 0;
         this.holderSerialNumber = BigInteger.ZERO;
+        this.holderIssuer = null;
     }
 
     /**
@@ -302,8 +304,10 @@ public abstract class Certificate extends ArchivableEntity {
                                             .getBaseCertificateID()
                                             .getSerial()
                                             .getValue();
+                this.holderIssuer = attCertInfo.getHolder()
+                        .getBaseCertificateID().getIssuer()
+                        .getNames()[0].getName().toString();
                 this.signature = attCert.getSignatureValue().getBytes();
-
                 this.issuer = getAttributeCertificateIssuerNames(
                                         attCertInfo.getIssuer())[0].toString();
                 this.issuerOrganization = getOrganization(this.issuer);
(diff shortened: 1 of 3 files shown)


HIRS_Utils/src/main/java/hirs/data/persist/certificate/Certificate.java
```diff
@@ -302,8 +304,10 @@ public Certificate(final byte[] certificateBytes) throws IOException {
                                             .getBaseCertificateID()
                                             .getSerial()
                                             .getValue();
+                this.holderIssuer = attCertInfo.getHolder()
+                        .getBaseCertificateID().getIssuer()
+                        .getNames()[0].getName().toString();
```
apldev3: Can't help but wonder if not checking for the presence of these fields and just assuming they'll be there ever causes significant issues.

cyrus-dev: I was following the style of code surrounding it.  I'm fine adding a null check on this line but then can we open up a ticket/milestone to parse the code base and correct the other locations?

iadgovuser19: I can't say anything bad about doing null checks -- always a great idea.

Something that should be considered here is the difference between the options available within RFC 5755 attribute certificate's holder field and the TCG Platform Certificate's holder field. All attribute certificates must have a holder field, however 5755 has three options of what can go in to a holder field. A Platform Certificate MUST only use the BaseCertificateID, which will always have issuer and serial. If there is no GeneralNames, you may have a legal 5755 attribute certificate, but you don't have a legal Platform Certificate. The issuer is of type "GeneralNames," which is a sequence of type "GeneralName" with a minimum size of 1 ("GeneralNames ::= SEQUENCE SIZE (1..MAX) OF GeneralName"). The .getNames() function returns an array of GeneralName. If that array does not have at least one entry, it is not legal ASN.1, and definitely not a legal Platform Certificate. The [0] element must be of type GeneralName, for which the getName() and toString() functions should always work. However...

Something that has a lot of room for improvement is how we deal with the GeneralName. The GeneralName can be any one of many, many types. We could/should run the getType() function to see what type of name it is before trying to parse it. It should always be an X500Name type (option '4'), because that is what the "issuer" field in an X.509v3 certificate is, which is what should be replicated into the holder of a Platform Certificate. Then, when we do the getName(), we know that the returning type can be cast as an X500Name. It is very tempting to immediately do a toString() to the X500Name, but this is not the best idea. An X500Name is a complex structure, and is subject to different interpretations when trying to read it as a string. We are using BouncyCastle when getting the string representation of the X500Name in the holder, but we are using the java.security representation when doing the same to the issuer field. Although these have always returned the same string in the past, there is no guarantee that they will always be the same. They should however be an identical ASN.1 structure, but that structure includes both SEQUENCEs (ordered sets) and SETs (not-necessarily-ordered sets). Two "identical" X500Name structures could theoretically be different sequences of bytes, with different string representations. We can either hope that this will never be different, or guarantee that we validate the similarity properly. This is just as important as null checking, IMHO.

apldev3: Alright, so it sounds to me like there are two issues that should be written up. First, we definitely should handle the case of doing proper null checking. Not sure what the best response is if a null is detected at this point, but we should start doing at least a check on the basis of us just passing in a byte array.

There's is nothing preventing that sequence of bytes from being incredibly unstructured, and we just always assume that it is properly formatted.

Furthermore, we should do better parsing of X500Name objects per @iadgovuser19.

If both of these are marked down as issues, I feel comfortable closing/merging this pull request with no additional work necessary.

Issue #21, opened 2018-10-05
ACA provsioning Platform Attributes pass when platform cert not found.


iadgovuser26:
When testing the platform validation process found a condition where the Platform Cert wasn't loaded but the Attributes seemed to pass.

Another condition where this possible is if the Platform cert was loaded but the signature was invalid (e.g. the Platform Certs Trust Chain is missing) then this case could be possible. The error indicated by hovering over the Platform cert red circle indicated that "cant validate a platform cert without a trust store". 

Proposed solution: Platform Attributes should not pass if the Platform Certificate is invalid. This implies that the EK certificate needs to be valid as well.

apldev3:
MR #21 was merged. Closing this now.

PR #22, opened 2018-10-05
[#21] Make the Platform Attributes validation check match Platform validation

cyrus-dev:
This pull request forces the supply chain validation service to verify that the Platform Credential has a status of PASS.  If it does not, no matter the outcome of the Attributes validation, the status of the Attributes can not be PASS.  This has to do with the attributes passing with no platform certificate.

Fix for #21

### DIFF
--- a/HIRS_AttestationCA/src/main/java/hirs/attestationca/service/SupplyChainValidationServiceImpl.java
+++ b/HIRS_AttestationCA/src/main/java/hirs/attestationca/service/SupplyChainValidationServiceImpl.java
@@ -16,6 +16,7 @@ import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Set;
+import java.util.HashMap;
 import org.apache.logging.log4j.Level;
 import hirs.appraiser.Appraiser;
 import hirs.appraiser.SupplyChainAppraiser;
@@ -98,6 +99,7 @@ public class SupplyChainValidationServiceImpl implements SupplyChainValidationSe
         SupplyChainPolicy policy = (SupplyChainPolicy) policyManager.getDefaultPolicy(
                 supplyChainAppraiser);
         boolean acceptExpiredCerts = policy.isExpiredCertificateValidationEnabled();
+        HashMap<PlatformCredential, SupplyChainValidation> credentialMap = new HashMap<>();
 
         List<SupplyChainValidation> validations = new ArrayList<>();
 
@@ -128,10 +130,13 @@ public class SupplyChainValidationServiceImpl implements SupplyChainValidationSe
                 while (it.hasNext()) {
                     PlatformCredential pc = it.next();
                     KeyStore trustedCa = getCaChain(pc);
-                    validations.add(validatePlatformCredential(pc, trustedCa, acceptExpiredCerts));
+                    SupplyChainValidation platformScv = validatePlatformCredential(
+                            pc, trustedCa, acceptExpiredCerts);
+                    validations.add(platformScv);
                     if (null != pc) {
                         pc.setDevice(device);
                         this.certificateManager.update(pc);
+                        credentialMap.put(pc, platformScv);
                     }
                 }
             }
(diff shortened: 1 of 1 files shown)


HIRS_AttestationCA/src/main/java/hirs/attestationca/service/SupplyChainValidationServiceImpl.java
```diff
@@ -151,8 +153,24 @@ public SupplyChainValidationSummary validateSupplyChain(final EndorsementCredent
...
                     PlatformCredential pc = it.next();
-                    validations.add(validatePlatformCredentialAttributes(pc, device.getDeviceInfo(),
-                            ec));
+                    SupplyChainValidation attributeScv = validatePlatformCredentialAttributes(
+                            pc, device.getDeviceInfo(), ec);
+
+                    if (platformScv != null) {
+                        if (platformScv.getResult() == AppraisalStatus.Status.FAIL
+                                || platformScv.getResult() == AppraisalStatus.Status.ERROR) {
+                            if (attributeScv.getResult() ==  AppraisalStatus.Status.PASS) {
```
apldev3: Is this `attributeScv` guaranteed to be non-null if `platformScv` isn't?

apldev3:
This looks good pending the outcome of the one question I have.

HIRS_AttestationCA/src/main/java/hirs/attestationca/service/SupplyChainValidationServiceImpl.java
cyrus-dev: I'm not 100% certain, but I was just going back over this and from all indications validatePlaformCredentialAttributes via buildValidationRecord return the object SupplyChainValidation.  I'm fine either way.

apldev3: Actually, based on how this reads, I don't think this update will work in all instances. Having the single instance of `platformScv` won't work when we have multiple Platform Credentials to validate. The last PC provided by the Iterator in the first block will be the one with that sets `platformScv` object.

Instead, a mapping is needed from `PlatformCredential`s to `SupplyChainValidation`s. Then you as you iterate through the Platform Credential Attribute Supply Chain Validation loop that you have here, you can get the corresponding `scv` for the current Platform Credential and perform the check as you are.

HIRS_AttestationCA/src/main/java/hirs/attestationca/service/SupplyChainValidationServiceImpl.java
```diff
@@ -151,8 +153,24 @@ public SupplyChainValidationSummary validateSupplyChain(final EndorsementCredent
                 Iterator<PlatformCredential> it = pcs.iterator();
                 while (it.hasNext()) {
                     PlatformCredential pc = it.next();
-                    validations.add(validatePlatformCredentialAttributes(pc, device.getDeviceInfo(),
-                            ec));
+                    SupplyChainValidation attributeScv = validatePlatformCredentialAttributes(
+                            pc, device.getDeviceInfo(), ec);
+
+                    if (platformScv != null) {
```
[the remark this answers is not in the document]
cyrus-dev: Removing check and based on the agreement, resolving the first comment.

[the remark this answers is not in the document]
apldev3: I suppose it doesn't hurt to have it then, but I'm fairly certain you can't enable PcAttributeValidation unless PcValidation has been enabled previously, at least insofar as using only the ACA Portal is concerned. The only way I could see triggering this is direct manipulation at the DB level. All that being said, it's cheap to check for nulls so we may as well do it.

apldev3:
I checked `PlatformCredential` and it does have a functioning HashCode, so it shouldn't be an issue that we're throwing it in a `HashSet`. At this point, I'd say this is good to go in.

Issue #25, opened 2018-10-12
Improve ACA/Provisioner RESTful Exception Handling


apldev3:
The ACA presently sends nondescript 500 error messages to the client Provisioners when an error occurs during the provisioning process. These messages should be made more descriptive and parsed by the Provisioners appropriately to provide more helpful feedback to the user for troubleshooting.

PR #26, opened 2018-10-12
[#23] Update HIRS Utils and ACA to handle certificate padding

### LINKED ISSUE #23: EKCert processing fails when EKCert has extra bytes in allocated memory when stored in  NvRAM
Currently the provisioner and ACA expect the EKcert to be on TPM's NvRAM with no leading or trailing bytes within the space allocated for the EKcert storage. In one case the EKcert had several trailing bytes which caused the following events to occur: 

1. The TPM2 provisioner pulled the entire memory allocated for EKcert which included extra bytes of non used data appended to the EKcert.
1. The TP2 provisioner sent the entire blob to the ACA, but the ACA failed to parse the EKcert. The ACA sent back a 500 error to the provisioner.
1. When the EKcert was edited to remove the trailing bytes and uploaded to the ACA it had no effect on the process. The ACA did not search for a stored EKcert when the EKcert sent failed. 

Suggested improvements:

1. Update the ACA to process the length parameter of all sent certs (including the platform cert) and trim off any extra data.
1. If the ACA fails to process the EKcert the ACA should log an error and send back an error to the provisioner indicating the EKCert was malformed and could not be processed.

apldev3:
This merge request updates the HIRS Utils `Certificate` class to ensure that all Certificates are trimmed of any padding/unnecessary bytes placed before or after the end of the main certificate body.

Additionally, if a malformed certificate of this type is detected, then the ACA will log the occurrence.

Closes #23


apldev3:
There are a number of TODOs that will be handled as part of #27.

apldev3:
Alright, well #27 is done, I just want this to go through before I put up the MR for that. #25 is almost done and will be hopefully pushed in the next day or two.

HIRS_Utils/src/main/java/hirs/data/persist/certificate/Certificate.java
```diff
@@ -328,6 +331,47 @@ public Certificate(final byte[] certificateBytes) throws IOException {
...
+                if (i >= certificateBytes.length - 3) {
+                    throw new IllegalArgumentException(malformedCertStringBuilder
+                            .append(" Certificate is nothing more than ASN.1 Sequence.")
+                            .toString());
+                }
+                // Mark the start of the first ASN.1 Sequence / Certificate Body
+                certificateStart = i;
+
+                // Parse the length as the 2-bytes following the start of the ASN.1 Sequence
+                certificateLength = (certificateBytes[i + 2] & 0xFF) << 8;
```
apldev3: Since Java doesn't have a concept of an unsigned value, we have to coerce the byte value from certificateBytes[i + 2] up to integer to prevent a byte overflow occurring and ending up with a large negative value for the certificateLength.

HIRS_Utils/src/main/java/hirs/data/persist/certificate/Certificate.java
```diff
@@ -328,6 +331,47 @@ public Certificate(final byte[] certificateBytes) throws IOException {
...
     }
 
+    @SuppressWarnings("magicnumber")
+    private byte[] trimCertificate(final byte[] certificateBytes) {
+        int certificateStart = 0;
+        int certificateLength = 0;
+
+        StringBuilder malformedCertStringBuilder = new StringBuilder(MALFORMED_CERT_MESSAGE);
+        for (int i = 0; i < certificateBytes.length; i++) {
+            if (i == certificateBytes.length - 1) {
```
apldev3: Throw an error if we reach the last byte of the provided "certificateBytes" since no Certificate length field could be found.

HIRS_Utils/src/main/java/hirs/data/persist/certificate/Certificate.java
```diff
@@ -328,6 +331,47 @@ public Certificate(final byte[] certificateBytes) throws IOException {
...
+                throw new IllegalArgumentException(malformedCertStringBuilder
+                        .append(" No certificate length field could be found.").toString());
+            }
+
+            // Look for first ASN.1 Sequence marked by the two bytes (0x30) and (0x82)
+            if (certificateBytes[i] == 0x30 && certificateBytes[i + 1] == (byte) 0x82) {
+                if (i >= certificateBytes.length - 3) {
```
apldev3: If we find the ASN.1 sequence and length field but only have those as bytes, then all we have is an empty ASN.1 sequence, so we throw an exception.

HIRS_Utils/src/main/java/hirs/data/persist/certificate/EndorsementCredential.java
```diff
@@ -373,6 +373,8 @@ private void parseSequence(final ASN1Sequence seq, final boolean addToMapping,
                     revision.getValue());
             LOGGER.debug("Found TPM Spec:" + tpmSpecification.toString());
         } else if (addToMapping && key.equals(TPM_SECURITY_ASSERTIONS)) {
+            // TODO(apldev3): Update this block to properly parse TPM Security Assertions
```
apldev3: Ignore any TODOs you see here, this is taken care of in the next MR that will be put up after this one.

HIRS_Utils/src/main/java/hirs/data/persist/certificate/Certificate.java
```diff
@@ -328,6 +331,47 @@ public Certificate(final byte[] certificateBytes) throws IOException {
         this.certAndTypeHash = Objects.hash(certificateHash, getClass().getSimpleName());
     }
 
+    @SuppressWarnings("magicnumber")
+    private byte[] trimCertificate(final byte[] certificateBytes) {
+        int certificateStart = 0;
+        int certificateLength = 0;
```
apldev4: This method would be less cryptic and cleaner if you used a `ByteBuffer`.

HIRS_Utils/src/main/java/hirs/data/persist/certificate/Certificate.java
apldev4: `ByteBuffer` has a `getShort()` method that takes an index and returns a short. `Short.toUnsignedInt()` will convert the short to the unsigned int.  `ByteBuffer` has a method called `remaining()` to tell you how many bytes are remaining.

HIRS_Utils/src/main/java/hirs/data/persist/certificate/Certificate.java
apldev3: I'm not sure how this is less cryptic or cleaner than the one that's here. At least with the current version, the for loop is straightforward and you don't have to ponder where you're at in the ByteBuffer compared to simple array indexing.

Also, I ran the tests and they passed, so I'm assuming that this non-pushed version was correctly implemented.

image

HIRS_Utils/src/main/java/hirs/data/persist/certificate/Certificate.java
apldev3: If you look at the code for `Short.toUnsignedInt()` it does this same trick that we do here, just without the bit shifting.

HIRS_Utils/src/main/java/hirs/data/persist/certificate/Certificate.java
apldev3: Went ahead and pushed the work for shifting to a ByteBuffer. Let me know what you think.

HIRS_Utils/src/main/java/hirs/data/persist/certificate/Certificate.java
```diff
@@ -328,6 +332,53 @@ public Certificate(final byte[] certificateBytes) throws IOException {
...
+                // Parse the length as the 2-bytes following the start of the ASN.1 Sequence
+                certificateLength = Short.toUnsignedInt(
+                        certificateByteBuffer.getShort(currentPosition + 2));
+                // Add the 4 bytes that comprise the start of the ASN.1 Sequence and the length
+                certificateLength += 4;
+                break;
+            }
+        }
+
+        if (certificateStart + certificateLength > certificateBytes.length) {
```
apldev4: Do you want to handle the case here where you never found the ASN.1 Sequence and `certificateStart + certificateLength` is 0?

apldev3: That case can't happen due to the first `if` statement under the `while` loop. It'll throw an `IllegalArgumentException` and this block is never reached.

Issue #28, opened 2018-10-17
SELinux Policy Issues with HIRS ACA on CentOS 7


m-morrone:
After installing ACA on a CentOS 7 system, navigating to the ACA Portal URL (https://<ACAserveraddress>:8443/HIRS_AttestationCAPortal/) displayed a blank page.  Looking in the tomcat logs, shows:
`192.168.103.1 - - [16/Oct/2018:19:04:20 +0000] "GET /HIRS_AttestationCAPortal HTTP/1.1" 404 -`
SELinux was blocking the connection from tomcat to mysqld, indicated in the audit logs by: `node=server22.simp.test type=AVC msg=audit(1539714234.510:4782): avc:  denied  { name_connect } for  pid=12058 comm="java" dest=3306 scontext=system_u:system_r:tomcat_t:s0 tcontext=system_u:object_r:mysqld_port_t:s0 tclass=tcp_socket`
Could you add a SELinux policy to the RPM to enable this out of the box?

apldev4:
The AttestationCA RPM contains the SELinux policy to fix this very issue. After installing the RPM, look for `/opt/hirs/extras/aca/tomcat-mysql-hirs.pp`. The loading of an SELinux policy by an RPM post-install script may not go over well with System Administrators who want to know what SELinux policies are being run, so the loading of the provided policy has been left to the System Administrator. Simply run the following command as root:
```
semodule -i /opt/hirs/extras/aca/tomcat-mysql-hirs.pp
```
You will need to make sure you have installed the `policycoreutils` package in order to have `semodule`.

m-morrone:
I couldn't find this documented.  Could you add it to the wiki at https://github.com/nsacyber/HIRS/wiki/aca_debug#no-attestation-portal-appears-at-hirs_server-package-portal-url?  Thanks.

iadgovuser26:
Agreed. This should probably be a note on the installation_notes page. It will, however, take a few days to update the pages.

trevor-vaughan:
@apldev1 I would recommend having the RPM compile the policy from scratch like we do over in https://github.com/simp/simp-environment-skeleton/tree/master/build.

As a user, I expect that things are just dropped into place. However, I would be just as uncomfortable running a randomly compiled policy file as I would having the RPM do it. I think having the source components available would make things a lot easier all around.

trevor-vaughan:
[...]

PR #29, opened 2018-10-17
[#28] ACA RPM modifies SELinux policy to allow Tomcat to use MySQL.

apldev4:
The CentOS7 package selinux-policy-targeted does not allow Tomcat
to use port 3306, which is the default MySQL port. This commit
changes the ACA RPM to modify the SELinux policy to grant that
permission on fresh installs. This makes the ACA RPM now require
the policycoreutils package to be installed.

### DIFF
--- a/package/rpm/HIRS.spec
+++ b/package/rpm/HIRS.spec
@@ -137,7 +137,7 @@ Prefix          : /usr/share/tomcat6
 %endif
 
 %if 0%{?build7}
-Requires        : mariadb-server, openssl, tomcat, java-1.8.0, rpmdevtools, coreutils, initscripts, chkconfig, sed, grep, firewalld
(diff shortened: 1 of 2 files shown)


package/rpm/HIRS.spec
```diff
@@ -164,6 +164,9 @@ if [ "$1" = "1" ]; then
     # open necessary ports
     sh /opt/hirs/scripts/common/firewall_configure_tomcat.sh
 
+    # Allow Tomcat to use port 3306 to communicate with MySQL
+    semodule -i /opt/hirs/extras/aca/tomcat-mysql-hirs.pp
```
apldev3: Based off discussion on phone call, perhaps we should have some check for SELinux enabled before running this command.

apldev4: I update this and tested it.

apldev3:
Wonder why this didn't seem to trigger a Travis CI build?

apldev3:
@apldev1's comment was fixed. Presumably that renders this merge request good in their regard.

PR #30, opened 2018-10-17
Add libtss2 support for Deb packaging.

apldev2:
Changes how tpm20.h, which contains the TPM2 SAPI, is imported to allow successful packaging of both debs and rpms. This is a first step to getting full support for Ubuntu, including potentially 16.04. It does not, however, complete the process.


apldev3:
LGTM as well.

PR #34, opened 2018-10-19
[#33] IMA baselines can match measurements based solely on hashes

### LINKED ISSUE #33: Update IMA baselines to be able to match measurements solely based on hashes
IMA baselines are typically used to match measurement records against baseline records using both their paths and hashes.  In certain cases, it is useful to be able to match IMA measurements solely based on their hashes and not filenames.  ImaAcceptableRecordBaseline and its subclasses should be updated to include a `containsHashes` which performs this operation.

apldev1:
ImaAcceptableRecordBaseline and its subclasses have been updated to include
a containsHashes method to be able to match IMA measurement records
based solely on their hashes.  Supporting classes have been
updated or created as necessary.

Additionally, the set of path equivalencies as specified in the IMA
policy have been updated to include additional entries.

Closes #33.

### DIFF
--- a/HIRS_Utils/src/main/java/hirs/data/persist/IMAPolicy.java
+++ b/HIRS_Utils/src/main/java/hirs/data/persist/IMAPolicy.java
@@ -457,9 +457,11 @@ public class IMAPolicy extends Policy implements HasBaselines {
         Multimap<String, String> equivalentPaths = HashMultimap.create();
 
         // define equivalences
-        equivalentPaths.put("/bin/", "/usr/bin/");
-        equivalentPaths.put("/lib/", "/usr/lib/");
-        equivalentPaths.put("/lib64/", "/usr/lib64/");
+        equivalentPaths.put("/bin/",     "/usr/bin/");
+        equivalentPaths.put("/lib/",     "/usr/lib/");
+        equivalentPaths.put("/lib64/",   "/usr/lib64/");
+        equivalentPaths.put("/usr/bin/", "/usr/sbin/");
+        equivalentPaths.put("/sbin/",    "/usr/sbin/");
 
         // populate inverse relationships
         Multimap<String, String> bidirectionalEquivalences = HashMultimap.create();
--- a/HIRS_Utils/src/test/java/hirs/data/persist/SimpleImaBaselineTest.java
+++ b/HIRS_Utils/src/test/java/hirs/data/persist/SimpleImaBaselineTest.java
@@ -1,5 +1,6 @@
 package hirs.data.persist;
 
+import hirs.ima.matching.BatchImaMatchStatus;
 import hirs.ima.matching.IMAMatchStatus;
 import hirs.persist.BaselineManager;
 import hirs.persist.DBBaselineManager;
@@ -221,6 +222,15 @@ public class SimpleImaBaselineTest extends SpringPersistenceTest {
         return baseline.contains(
                 Collections.singletonList(record), recordManager, imaPolicy
         ).getIMAMatchStatuses(record).iterator().next();
+    }
+
+    private BatchImaMatchStatus<IMABaselineRecord> baselineContainsHashes(
+            final SimpleImaBaseline baseline,
+            final IMAMeasurementRecord record,
+            final IMAPolicy imaPolicy) {
+        return baseline.containsHashes(
+                Collections.singletonList(record), recordManager, imaPolicy
+        );
     }
 
     /**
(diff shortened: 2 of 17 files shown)


HIRS_Utils/src/main/java/hirs/data/persist/IMAPolicy.java
```diff
@@ -457,9 +457,11 @@ public final void setPartialPathEnable(final boolean partialPathEnable) {
...
 
         // define equivalences
-        equivalentPaths.put("/bin/", "/usr/bin/");
-        equivalentPaths.put("/lib/", "/usr/lib/");
-        equivalentPaths.put("/lib64/", "/usr/lib64/");
+        equivalentPaths.put("/bin/",     "/usr/bin/");
+        equivalentPaths.put("/lib/",     "/usr/lib/");
+        equivalentPaths.put("/lib64/",   "/usr/lib64/");
+        equivalentPaths.put("/usr/bin/", "/usr/sbin/");
+        equivalentPaths.put("/sbin/",    "/usr/sbin/");
```
iadgovuser26: Does one still need to do equivalent paths if check is done by hash? Is there some benefit to leaving this in such as number of db accesses?

HIRS_Utils/src/test/java/hirs/ima/matching/ImaAcceptableHashRecordMatcherTest.java
```diff
@@ -0,0 +1,153 @@
...
+                            new IMABaselineRecord(FILENAME_1, HASH_1),
+                            new IMABaselineRecord(FILENAME_2, HASH_1))
+                        ),
+                        baseline
+                )
+        );
+    }
+
+    /**
+     * Tests that contains functions if a matching record is given in the case of that matching
```
apldev4: This javadoc does not describe what the test is doing.

HIRS_Utils/src/test/java/hirs/ima/matching/ImaAcceptableHashRecordMatcherTest.java
```diff
@@ -0,0 +1,153 @@
...
+public class ImaAcceptableHashRecordMatcherTest {
+    private static final String FILENAME_1 = "/bin/ls";
+    private static final String FILENAME_2 = "/bin/ls_with_another_name";
+    private static final Digest HASH_1 =
+            SimpleImaBaselineTest.getDigest("33333c2f7f3003d2e4baddc46ed4763a49543333");
+    private static final Digest HASH_2 =
+            SimpleImaBaselineTest.getDigest("00000c2f7f3003d2e4baddc46ed4763a49543333");
+
+    /**
+     * Tests that contains functions if no records are given.
```
apldev4: All of the test function javadocs begin with "Tests that contains", which is a bit confusing English until you realize that the word "contains" in the comment refers to a method. Maybe put some parenthesis after the word contains to be clearer.

HIRS_Utils/src/test/java/hirs/ima/matching/ImaAcceptablePathAndHashRecordMatcherTest.java
```diff
@@ -99,16 +100,65 @@ public void testContainsEquivalentPaths() {
...
+        List<IMAMeasurementRecord> measurementRecords = Arrays.asList(
+                new IMAMeasurementRecord("/sbin/dhclient", ONES),
+                new IMAMeasurementRecord("/sbin/sysctl", ONES),
+                new IMAMeasurementRecord("/sbin/swapon", ONES),
+                new IMAMeasurementRecord("/usr/sbin/audispd", ONES),
+                new IMAMeasurementRecord("/sbin/sysctl", ONES),
+                new IMAMeasurementRecord("/usr/sbin/ldconfig", ONES),
+                new IMAMeasurementRecord("/usr/sbin/kexec", ONES),
+                new IMAMeasurementRecord("/sbin/ip", ONES),
+                new IMAMeasurementRecord("/bin/bash", ONES)
```
apldev4: This tests the equivalence of `/sbin` to `/usr/sbin` and the reverse several times. It also tests the equivalence of `/usr/bin` to `/bin`, but not its reverse. Of the 10 directional equivalences, only 3 are being tested. There are actually more than 10 that should be tested since `/bin` is equivalent to `/usr/bin` and `/usr/bin` is equivalent to `/usr/sbin`, which is equivalent to `/sbin`.

HIRS_Utils/src/main/java/hirs/data/persist/IMAPolicy.java
apldev4: I think these are still needed for the case where path and hash both need to "match"--meaning equivalent. However, there are some transitive equivalencies that are not captured here. For example, `/bin` is equivalent to `/usr/bin` which is equivalent to `/usr/sbin` which is equivalent to `/sbin`. Perhaps a list of equivalent sets would be easier to manage.

apldev1: Yes, these are still needed because they still inform matches & mismatches.  Checking by hash is only done for records that are otherwise unknown.

Although the term 'equivalent' would suggest these paths should be transitive, we've only been adding explicit equivalent pairs so far, and not making these transitive to more tightly capture the actual installation locations we're seeing in RPMs.

HIRS_Utils/src/test/java/hirs/data/persist/SimpleImaBaselineTest.java
```diff
@@ -601,6 +611,74 @@ public final void containsEquivalentPath() {
...
+        );
+
+        baseline.addToBaseline(gradleBaselineRecord);
+        baseline.addToBaseline(new IMABaselineRecord(
+                baselineGradleFilename,
+                getDigest("00000c2f7f3003d2e4baddc46ed4763a49543333")
+        ));
+
+        IMAMeasurementRecord record = new IMAMeasurementRecord(
+                measuredGradleFilename, gradleHash
```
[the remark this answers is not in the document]
apldev1: This test is meant to ensure that there is a match between a measurement record with `measuredGradleFilename` and a baseline record with the different `baselineGradleFilename`.  Both records in the baseline use `baselineGradleFilename`, but only one of them matches, based on the hash.  I'll expand the test to include all four permutations of filename and hash, and that only the expected two are returned.

HIRS_Utils/src/test/java/hirs/ima/matching/ImaAcceptableHashRecordMatcherTest.java
apldev1: Thanks, will fix.

HIRS_Utils/src/test/java/hirs/ima/matching/ImaAcceptableHashRecordMatcherTest.java
apldev1: Will do.

HIRS_Utils/src/test/java/hirs/ima/matching/ImaAcceptablePathAndHashRecordMatcherTest.java
apldev1: Thanks!  I based this unit test off some real-world examples we saw, but I will expand the cases to handle all 10 current directional equivalences.  At this time, we are not considering these equivalencies as transitive, so there are indeed 10 cases to check.

apldev1:
@apl4 I've addressed your comments, please see if my changes resolve your concerns!

HIRS_Utils/src/test/java/hirs/data/persist/SimpleImaBaselineTest.java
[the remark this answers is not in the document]
apldev4: Looks good.

HIRS_Utils/src/test/java/hirs/ima/matching/ImaAcceptableHashRecordMatcherTest.java
apldev4: looks good.

HIRS_Utils/src/test/java/hirs/ima/matching/ImaAcceptableHashRecordMatcherTest.java
apldev4: ok

HIRS_Utils/src/test/java/hirs/ima/matching/ImaAcceptablePathAndHashRecordMatcherTest.java
apldev4: Very good.

PR #35, opened 2018-10-23
[#7] Ensure Ubuntu support pending end-user installation of supported…

### LINKED ISSUE #7: Feature request: HIRS support Ubuntu 18.04 LTS

apldev3:
… TPM2 Libraries

This merge request will be paired with a brief write-up to be added to the wiki that walks the end-user through installing a supported version of Tpm2-Tss and Tpm2-Tools on Ubuntu in light of the fact that the core package repositories of most current Ubuntu versions do not presently have a HIRS-supported version of Tpm2-Tss and/or Tpm2-Tools available for simple `apt-get` installation.

This merge request removes DEB package dependencies for both of these libraries since they're not guaranteed to be valid versions.

Additionally, this work also opens up TPM2 Provisioning support to anyone running a Tpm2-Tools Version 3.X.X since they've started stabilizing their releases and general functionality around this Major Version.

Closes #7


HIRS_ProvisionerTPM2/test/Utils_Test.cpp
```diff
@@ -536,6 +536,26 @@ TEST_F(UtilsTest, ParseTpm2ToolsVersionFailNonSemanticVersion) {
...
+TEST_F(UtilsTest, ParseTpm2ToolsMajorVersionSuccessMultiNumeralVersion) {
+    stringstream versionOutput;
+    versionOutput << "10.29.970";
+
+    string expectedOutput = "10";
+    string majorVersion = Tpm2ToolsOutputParser::parseTpm2ToolsMajorVersion(
+            versionOutput.str());
+    ASSERT_EQ(expectedOutput, majorVersion);
+}
+
```
apldev1: Could you add a test here that ensures `parseTpm2ToolsMajorVersion` returns an empty string if it can't parse its argument?

PR #37, opened 2018-10-24
[#36] TPM Provisioner status output

### LINKED ISSUE #36: TPM Provisioning is missing final status
Some status has been inadvertently removed from the hirs-provision process
This is what is currently displayed from a tpm_aca_provision:
```
TPM 1.2 detected.
--> Configuring provisioner
----> Deleting existing key store
----> Downloading truststore
--> Provisioning
----> Removing old provisioner scripts, if any
----> Provisioning TPM
```
An what used to be displayed
```
TPM 1.2 detected.
--> Configuring provisioner
----> Deleting existing key store
----> Downloading truststore
--> Provisioning
----> Provisioning TPM
----> Collecting Device Information
----> Got Endorsement Credential from TPM
----> Sending Attestation Identity Credential Request
----> Attestation Identity Provisioning succeeded 
----> TPM successfully Provisioned
```

mtsai1:
This change captures the log output from HIRS_Provisioner in the console like so:
```
TPM 1.2 detected.
--> Configuring provisioner
----> Deleting existing key store
----> Downloading truststore
--> Provisioning
----> Removing old provisioner scripts, if any
----> Provisioning TPM
----> Collecting device information
----> Got endorsement credential from TPM
----> Sending Attestation Identity Credential Request
----> Attestation Identity Provisioning succeeded
```

Closes #36

### DIFF
--- a/HIRS_Provisioner/src/main/java/hirs/provisioner/client/RestfulClientProvisioner.java
+++ b/HIRS_Provisioner/src/main/java/hirs/provisioner/client/RestfulClientProvisioner.java
@@ -135,14 +135,14 @@ public class RestfulClientProvisioner implements ClientProvisioner {
         LOG.debug(String.format("created TPM identity request using UUID %s", uuid));
 
         // attest the request with the ACA and obtain it's response
-        LOG.info("Sending Attestation Identity Credential Request");
+        System.out.println("----> Sending Attestation Identity Credential Request");
         IdentityResponseEnvelope response = attestIdentityRequest(identityRequest);
         if (null == response) {
             LOG.error("Provisioning failed.  Please refer to the Attestation CA for details.");
             return false;
         }
 
-        LOG.info("Attestation Identity Provisioning succeeded");
+        System.out.println("----> Attestation Identity Provisioning succeeded");
 
         // activate the identity with the TPM
         byte[] credential = activateCredential(response, uuid);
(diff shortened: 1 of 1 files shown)


HIRS_Provisioner/src/main/resources/log4j2.xml
```diff
@@ -16,7 +16,7 @@
         </RollingFile>
     </Appenders>
     <Loggers>
-        <logger name="hirs" level="WARN" additivity="false">
+        <logger name="hirs" level="INFO" additivity="false">
```
apldev3: Considering that this would potentially enable logging of other INFO statements, perhaps the better idea would be to simply do a standard `System.out.println` with the usual `---->` prepended to the statement? That's how they seem to be handled in the corresponding shell scripts as `echo "-->Provisioning"`

What do you think @apldev4 or @apldev1

mtsai1: Any thoughts?  I agree that changing the specific log.info() statements to System.out.println would probably be better in the long run than broadly changing the log level.

apldev3: Yeah I would suggest going ahead and making the change to `System.out.println`.

apldev3:
@iadgovuser26 does this update look fine to you?

PR #39, opened 2018-10-29
[#25] Make ACA exception handling more descriptive

apldev3:
This merge request updates the ACA to return errors to the Client Provisioner as a JSON-serialized POJO that contains a simple error message that can/will be logged on the client side. While the messages themselves may need fine-tuned overtime the architecture here should let us add new exception messages/improve information returned by the ACA to the client in future tickets.

Closes #25

PR #40, opened 2018-10-30
 [#38] ACA checks uploaded EK Certs if one is not provided during prov…

### LINKED ISSUE #38: ACA Should Retrieve Uploaded EK Certificates for TPM 2.0 Provisioning
While working #25 it has been discovered that the ACA is not retrieving uploaded EK Certs during TPM 2.0 Provisioning. It will look for uploaded Platform Certificates, but the equivalent check for Endorsement Certificates is not performed. This should be fixed such that the TPM 2.0 Provisioning process behaves like the old TPM 1.2 Provisioning process.

apldev3:
...isioning

This MR requires #39 to be merged before this should go in. Once it does, this is actually a very small merge request to ensure that the ACA checks for any uploaded EK Certificates when performing TPM 2.0 Provisioning. In it's current state, it would check for uploaded Platform Certificates, but not uploaded EK Certificates, which does not keep in line with the original TPM 1.2 Provisioner.

Closes #38

### DIFF
--- a/HIRS_AttestationCA/src/main/java/hirs/attestationca/AbstractAttestationCertificateAuthority.java
+++ b/HIRS_AttestationCA/src/main/java/hirs/attestationca/AbstractAttestationCertificateAuthority.java
@@ -206,7 +206,8 @@ public abstract class AbstractAttestationCertificateAuthority
     @Override
     public byte[] processIdentityRequest(final byte[] identityRequest) {
         if (ArrayUtils.isEmpty(identityRequest)) {
-            throw new IllegalArgumentException("identityRequest cannot be null or empty");
+            throw new IllegalArgumentException("The IdentityRequest sent by the client"
+                    + " cannot be null or empty.");
         }
 
         LOG.debug("received request to process identity request");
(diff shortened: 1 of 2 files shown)


HIRS_AttestationCA/src/main/java/hirs/attestationca/AbstractAttestationCertificateAuthority.java
```diff
@@ -1366,21 +1379,29 @@ private TPM2ProvisionerState getTpm2ProvisionerState(
...
+        EndorsementCredential endorsementCredential = null;
         if (identityClaim.hasEndorsementCredential()) {
-            return CredentialManagementHelper.storeEndorsementCredential(
+            endorsementCredential = CredentialManagementHelper.storeEndorsementCredential(
                     this.certificateManager,
                     identityClaim.getEndorsementCredential().toByteArray());
+        } else if (ekPub != null) {
+            LOG.warn("Endorement Cred was not in the identity claim from the client."
+                    + " Checking for uploads.");
+            endorsementCredential = getEndorsementCredential(ekPub);
```
iadgovuser26: What happens if no pubek was sent? Seems like that would result in an error somewhere else. Should that be logged?

iadgovuser26: What happens if no pubek was sent?  Should that be checked for and logged?

apldev3: I'll answer your one question since I'm pretty sure they're the same. Namely, that can't happen in the current form. The IdentityClaim object is required to have an EkPublicArea so there has to be some amount of bytes there. The PublicKey itself is parsed before this gets called and if it was malformed an error would be thrown before we get to this point.

apldev3: See other comment.

HIRS_AttestationCA/src/main/java/hirs/attestationca/AbstractAttestationCertificateAuthority.java
```diff
@@ -425,12 +430,13 @@ private IdentityResponseEnvelope generateIdentityResponseEnvelopeAndStoreIssuedC
      * Performs supply chain validation.
      *
      * @param claim the identity claim
+     * @param ekPub the public endorsement key
      * @return the {@link AppraisalStatus} of the supply chain validation
      */
     private AppraisalStatus.Status doSupplyChainValidation(
-            final ProvisionerTpm2.IdentityClaim claim) {
+            final ProvisionerTpm2.IdentityClaim claim, final PublicKey ekPub) {
```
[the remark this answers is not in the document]
apldev3: Yeah, trust me, I had considered that route, but we also use the ekPub right after this method completes in the calling method, thus instead of regenerating the key twice and wasting cycles, I figured we could generate it once in the calling method and use it in both places.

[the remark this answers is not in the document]
apldev1: Fair enough!

Issue #41, opened 2018-10-30
Provisioners should use PACCOR for device info collection


apldev4:
Instead of having device info collection implemented twice -- once in Java from 1.2 Provisioning and once in C++ for 2.0 Provisioning, just use PACCOR to collect the device info.

apldev3:
The MR for this issue has been merged and this issue can be closed.

PR #45, opened 2018-11-02
[#41] Provisioners use PACCOR for device info collection.

apldev4:
The provisioners used to shell out using different tools
to collect device info. Now they both use PACCOR instead.


HIRS_ProvisionerTPM2/docs/dependencies-centos.md
```diff
@@ -13,5 +13,4 @@ Please look up their respective names in the CentOS repositories.
 | protobuf   | 2.5.0        | 2.4.1 (estimated) | CentOS 7 base         | https://github.com/google/protobuf          |
 | re2        | 20160401     | 20160201          | CentOS 7 epel-release | https://github.com/google/re2               |
 | tpm2-tss   | 1.2.0        | 1.0.0             | CentOS 7 base         | https://github.com/intel/tpm2-tss           |
-| lshw       | B.02.18-12   | B.02.18-12        | CentOS 7 base         | https://github.com/lyonel/lshw              |
```
apldev3: We'll need to remember to remove this dependency from the different Guides in the Wiki when this gets merged.

HIRS_ProvisionerTPM2/src/TPM2_Provisioner.cpp
```diff
@@ -70,6 +71,9 @@ int provision() {
                                        endorsementCredential,
                                        platformCredentials);
     identityClaim.set_client_version(CLIENT_VERSION);
+    string paccorOutputString =
+            RUN_PROCESS_OR_THROW("/opt/paccor/scripts/allcomponents.sh", "");
+    identityClaim.mutable_paccoroutput()->assign(paccorOutputString);
```
apldev3: Should be able to use `identityClaim.set_paccoroutput(paccorOutputString);`

HIRS_ProvisionerTPM2/docs/dependencies-centos.md
apldev3: Actually for that matter does paccor need added to this list?

package/rpm/HIRS.spec
```diff
@@ -36,11 +36,11 @@ Summary         : Host Integrity at Runtime and Startup (HIRS) Provisioner
 Group           : System Environment/Base
 
 %if 0%{?build6}
-Requires        : tpm_module, java-1.8.0, wget, util-linux, chkconfig, sed, initscripts, coreutils, dmidecode, lshw, bash%{?RPM_EXTRA_CLIENT_DEPENDENCIES}
+Requires        : tpm_module, java-1.8.0, wget, util-linux, chkconfig, sed, initscripts, coreutils, dmidecode, paccor, bash%{?RPM_EXTRA_CLIENT_DEPENDENCIES}
```
apldev3: This dependency will be fine if you build Paccor from source and install it, correct?

HIRS_ProvisionerTPM2/CMakeLists.txt
```diff
@@ -257,7 +257,7 @@ if (${DISTRIBUTION} STREQUAL "Ubuntu")
     set(CPACK_GENERATOR "DEB")
     set(CPACK_DEBIAN_PACKAGE_NAME "HIRSProvisionerTPM2.0")
     set(CPACK_DEBIAN_PACKAGE_SECTION "admin")
-    set(CPACK_DEBIAN_PACKAGE_DEPENDS "liblog4cplus-1.1-9(>=1.1.2), libcurlpp0(>=0.7), lshw")
```
apldev3: Should paccor be added here?

HIRS_ProvisionerTPM2/CMakeLists.txt
```diff
@@ -279,7 +279,7 @@ elseif (${DISTRIBUTION} STREQUAL "CentOS Linux")
     set(CPACK_RPM_PACKAGE_RELEASE_DIST "el7")
     set(CPACK_RPM_PACKAGE_LICENSE "Apache License, Version 2.0")
     set(CPACK_RPM_PACKAGE_GROUP "System Environment/Base")
-    set(CPACK_RPM_PACKAGE_REQUIRES "log4cplus >= 1.1.2, tpm2-tss >= 1.0, tpm2-tools >= 1.1.0, protobuf >= 2.4.1, re2 >= 20160401, libcurl >= 7.0.0, lshw")
```
apldev3: And also paccor added here?

HIRS_ProvisionerTPM2/docs/dependencies-centos.md
apldev4: I will add it to this file.

package/rpm/HIRS.spec
apldev4: No. If you install something outside of the realm of the package manager (in this case RPM), the package management will not know about it, so dependencies cannot be confirmed. However, the installer can do an install that ignores dependencies.

HIRS_ProvisionerTPM2/CMakeLists.txt
apldev4: Added it.

HIRS_ProvisionerTPM2/CMakeLists.txt
apldev4: Added it.

HIRS_ProvisionerTPM2/src/TPM2_Provisioner.cpp
apldev4: changed.

package/rpm/HIRS.spec
apldev3: So if that's the case do we want to list paccor as a requires? It's not currently listed in any package repos and would then force the user to always do an install and ignore dependencies, which seems suspect.

HIRS_ProvisionerTPM2/CMakeLists.txt
apldev3: Let's hold up for a second here. We should probably come to some sort of consensus regarding including paccor as a depedency in the RPMs and DEBs. If the user would have to ignore dependencies to install the package just for the sake of Paccor, then it shouldn't be included. More likely we should write a separate issue to detect Paccor (similar to the issue to detect TPM2-Abrmd).

HIRS_ProvisionerTPM2/CMakeLists.txt
apldev3: See above comment.

HIRS_ProvisionerTPM2/docs/dependencies-centos.md
apldev3: I would say Paccor is fine here regardless of the discussion regarding the Package Requirements.

package/rpm/HIRS.spec
apldev4: By install, I thought you meant "make install" on source code. If the user installs PACCOR via RPM, everything is fine. Yes, we definitely want to list PACCOR as a dependency because the Provisioners don't work at all without it.

apldev3: Well I did mean `make install` but if they can just install the RPM and that registers it with the Package Manager such that it won't block the installation of the Provisioner, then this is fine with me. As a final check, that's what you're suggesting?

HIRS_ProvisionerTPM2/CMakeLists.txt
apldev3: Pending the answer to my last question, this should be fine.

package/rpm/HIRS.spec
apldev4: Yes.

apldev3:
Everything looks good to me at this point.

apldev4:
I can update the Install Guide.

HIRS_Utils/src/main/java/hirs/client/collector/DeviceInfoCollector.java
```diff
@@ -100,6 +105,66 @@ public DeviceInfoCollector(
         LOGGER.warn("default constructor called! Should only be done in test");
     }
 
+    private static String callPACCOR() throws CollectorException {
```
apldev2: nit: I believe our style guide says we should only capitalize the first letter of acronyms in camel case

HIRS_Utils/src/main/java/hirs/validation/SupplyChainCredentialValidator.java
```diff
@@ -95,6 +98,42 @@ public SupplyChainCredentialValidator() {
 
     }
 
+    /**
+     * Parses the output from PACCOR's allcomponents.sh script into ComponentInfo objects.
+     * @param paccorOutput the output from PACCOR's allcomoponents.sh
+     * @return a list of ComponentInfo objects built from paccorOutput
+     * @throws IOException if something goes wrong parsing the JSON
+     */
+    public static List<ComponentInfo> getComponentInfoFromPACCOROutput(final String paccorOutput)
```
apldev2: nit: paccor should be in camel case

HIRS_Utils/src/main/java/hirs/client/collector/DeviceInfoCollector.java
apldev4: changed.

HIRS_Utils/src/main/java/hirs/validation/SupplyChainCredentialValidator.java
apldev4: changed.

PR #48, opened 2018-11-05
[#47] Prevent deletion of external dependencies for TPM 2.0 Provisioner

### LINKED ISSUE #47: Cache Downloaded Dependencies in TPM 2.0 Provisioner Build
Presently, both Gradle and the packaging scripts for the TPM 2.0 Provisioner perform a simple delete of the build directory anytime you attempt to build or package this particular subproject. This creates an issue in environments where Internet service is spotty due to the deletion of downloaded external dependencies.

While these dependencies cannot be removed, we can avoid deleting them to at least "cache" the dependencies.

Both the building and packaging scripts should be updated to prevent deletion of these external dependencies so as to prevent having to redownload them everytime the package requires packaging or a Gradle build.

apldev3:
This merge request updates the Gradle Build and TPM 2.0 Provisioner packaging scripts to avoid deleting the `lib` folder in the target build directory. In doing so, this effectively "caches" the external dependencies and allows for CMake to detect and make use of the dependencies that were already downloaded.

Closes #47

### DIFF
--- a/HIRS_ProvisionerTPM2/build.gradle
+++ b/HIRS_ProvisionerTPM2/build.gradle
@@ -1,10 +1,5 @@
-task cleanup() {
-    delete 'build'
-}
-
-task prep(type: Exec) {
-    workingDir './'
-    commandLine 'mkdir', 'build'
+task cleanup(type: Delete) {
+	delete fileTree(dir: 'build', exclude: 'lib/')
 }
 
 task cmake(type: Exec) {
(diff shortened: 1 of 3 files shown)


HIRS_ProvisionerTPM2/build.gradle
```diff
@@ -1,13 +1,22 @@
 task cleanup() {
-    delete 'build'
-}
-
-task prep(type: Exec) {
-    workingDir './'
-    commandLine 'mkdir', 'build'
+    delete 'build/CMakeCache.txt',
```
apldev4: Can you find a way to do this that does not require people to update this list of non`lib` subdirectories of `build`?

apldev3: Think I found a pretty good solution to this.

PR #53, opened 2018-11-14
[#52] Make TPM2 Provisioner check for a running Resource Manager

### LINKED ISSUE #52: TPM2 Provisioner Should Check for Resourcemgr or TPM2-Abrmd
Presently, the TPM 2.0 Provisioner relies on having the TPM2-Abrmd running, or for much older versions of the TPM2 Software Stack, the Resourcemgr daemon is required. Either way, a check should be added in at the start of provisioning to ensure that either of these daemons are running, otherwise we should abort the provisioning process and alert the user.

apldev3:
MR to make the TPM 2 Provisioner check if either `tpm2-abrmd` or `resourcemgr` is running locally before starting provisioning. Will warn the user if it's not detected.

In addition to this simple functionality, some refactoring was required. The `tpm2-tools-utils` section of `Utils` depended on the `Process` class, but to perform this update, `Process` would need to depend on `Utils`. Of course this created a cyclic dependency that need to be broken. To resolve the `tpm2-tools-utils` section of the `Utils` file was broken into its own file called `Tpm2ToolsUtils`. This new file could depend on `Process` and `Process` could depend on `Utils` without any problems.

All `CMakeLists.txt` files have been updated to reflect these changes.

Closes #52


apldev3:
Here's what the new message looks like when you run attempt to start provisioning. (Assuming you've run as a privileged user.)

image

HIRS_ProvisionerTPM2/src/Process.cpp
```diff
@@ -131,3 +141,62 @@ string Process::run(const string& executable,
...
 }
+
+/**
+ * Static function to check if a specified process is currently running in the
+ * local environment.
+ *
+ * @param executable the executable to check is running
+ * @return true, if executable is found to be running / false, otherwise
+ */
+bool Process::isRunning(const string& executable) {
```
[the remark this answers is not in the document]
apldev3: I like the `pgrep` suggestion and wouldn't bother using the others, but `pgrep` has its own problem in that it returns a non-zero error code when no processes are matched. Since we'd likely be using the `RUN_PROCESS_OR_THROW` macro, that would result in an Exception being thrown, which requires a `try/catch` for what could easily be handled by an `if/then` statement.

Additionally, upon checking the `man` page of the `pgrep` utility, it would appear that they are checking the `/proc/_pid_/stat` file unless you provide the `-f` flag that shifts it to use the `/proc/_pid_/cmdline` file. This is exactly what this function currently does without relying on yet another external utility. Rather than refactoring the `Process` class further beyond the scope of this ticket to prevent an Exception from being thrown for expected behavior. This seems to be a good enough solution than spend more time on it.

Additionally, a number of tests have been written testing this method. If there are any suggestions regarding further tests, I'm more than happy to write up additional test cases.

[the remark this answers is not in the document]
apldev4: 1. You don't have to use RUN_PROCESS_OR_THROW, especially since it does not make sense in this case.
2. I'd prefer to use tools that have some pedigree rather than trying to rewrite them because edge cases have been dealt with.
3. What is the advantage of searching `/proc/_pid_/cmdline` over `/proc/_pid_/stat`?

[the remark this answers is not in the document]
apldev3: 1. The only option I have since `isRunning` is a `static` method is `run`, which is just what `RUN_PROCESS_OR_THROW` wraps. Changing `isRunning` to be a non-static method seems like a poor design decision since you shouldn't need a `Process` object to determine if a system process is running when `Process` is a utility class of sorts. `isRunning` should belong to the class.
2. Typically I would agree, but in this case, this isn't an overtly complex scenario. The only truly odd edge case is if you're examining a specific process that terminates when you're about to check its `stat` or `cmdline` file. This would of course delete the file that you thought was there. For that case, there's a try/catch as the backup to ensure the program doesn't enter some strange realm due to this potential race condition. Instead it just continues onto the next process in the `/proc` file.
3. Per the `pgrep` utility's `man` page, the `cmdline` file is useful for situations where the process name may be longer than 15 characters. On some systems, this is a truncation point for the `stat` file, but the `cmdline` file will have the complete command that was run to execute the process. Due to the difference in content between the two files, `stat` is preferred for most cases as it has little to no chance of providing a false positive, whereas the `cmdline` increases the chance of a false positive, but is a necessary alternative failing `stat`.

HIRS_ProvisionerTPM2/src/Process.cpp
```diff
@@ -131,3 +142,68 @@ string Process::run(const string& executable,
...
+ * Static function to check if a specified process is currently running in the
+ * local environment.
+ *
+ * @param executable the executable to check is running
+ * @param checkCmdline a flag to check the complete cmdline commands
+ *  rather than just search for the process name, default false
+ * @return true, if executable is found to be running / false, otherwise
+ */
+bool Process::isRunning(const string& executable, bool checkCmdline) {
+    // If the executable is the empty string, return false
```
apldev4: You could replace the contents of this method with the following line and get the functionality you want. You can still have your static method so the code that uses it is easy to read. You just get less lines of code and the ability to rely on the fact that `pgrep` has had years of testing. Less code == less maintenance $$
```suggestion
    return Process("pgrep", "-c " + executable).run() == 0;
```

apldev3: In the case of this project, that last sentiment has burned us before, and my hesitance to bring in additional tooling to handle an otherwise simple task is reflective of that. That being said, we each clearly have our own amount of skepticism of either approach that is at a cross-purpose with one another. We'll go with your route on the basis of moving things along.

apldev3: I just experimented with the proposed `pgrep` solution and it doesn't work. Since we're invoking the tool `pgrep` via C++ and C++ (and likely C) uses the `sh` command to invoke commands via `popen`, then it will trivially always have a process running for the `cmdline` usage.

I'm putting the old version back in and this is being marked as resolved.

apldev3: image

Here's the failing test case demonstrating the aforementioned issue.

HIRS_ProvisionerTPM2/src/Process.cpp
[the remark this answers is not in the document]
apldev3: Per the other discussion, this approach does not work when support for `cmdline` is desired.

HIRS_ProvisionerTPM2/src/Process.cpp
apldev4: The new test above fails when testing a capability involving searching `/proc/PID/cmdline`. Neither `pgrep` nor your implementation of it correctly handles this file they way we want. The `/proc/PID/cmdline` file contains more than the executable name. It includes the fully qualified executable path and all of the arguments -- either of which could contain the character string you are receiving as `executable`. Furthermore, `isRunning()` currently passes the responsibility of deciding which value to use for `checkCmdline` to the user of the API along with the risks associated with passing `true` to the current implementation. There are two good fixes here:

1.  Since the two uses of `isRunning()` you need to complete this issue are both below the 15 character limit, they don't require you to write an implementation that correctly handles searching `/proc/PID/cmdline`. That means you can easily eliminate `checkCmdline` and its current logic, disallow `executable` from containing paths, and compare `executable` to `/proc/PID/stat`. `pgrep` can do the later for you using the single line I provided above -- even with C++ calling `pgrep` via `sh -c`. Either way, given that the 15 character limit is a function of the `stat` file and not of `pgrep` or your pgrep implementation, you should log a warning that `isRunning()` could be getting a false negative when `executable` is more than 15 characters and is being compared to `proc/PID/stat`.

2. If you really want to provide the additional capability to handle the case where `executable` is over 15 characters, I still recommend removing the `checkCmdline` parameter. If the 15 character limit comes into play, `isRunning()` should be able to determine if looking at `/proc/PID/cmdline` is necessary.  If it is necessary, you could modify your implementation of `pgrep` to take the first token of `/proc/PID/cmdline` after tokenizing on whitespace, extract the filename at the end of that token, and compare it to `executable`. This would provide an improvement over `pgrep` rather than just reimplementing it. If you are going to allow `executable` to contain absolute or relative paths, a full solution must deal with a lot more (e.g., differing but equivalent paths).

apldev3: Alright, so taking the first approach, it's still not quite as simple as suggested and there are still edge cases to consider regardless of this approach.

Specifically, consider the following.

image

If you are comfortable moving forward with this implementation, then I'll upload it, but otherwise, I'd rather not overwrite everything, and then have to switch back later.

My point here is that even this seemingly simplified approach of just running it as a Process has Edge Cases. It's whether you'd rather risk those edge cases as a race condition or as a potential Security liability.

apldev3: Alright after a quick outside discussion, I've pushed the above code as pictured. I also pushed the commit and we can see about leveraging Github's automatic rebasing capability to handle squashing all the commits into one.

HIRS_ProvisionerTPM2/CMakeLists.txt
```diff
@@ -234,7 +234,7 @@ if(STATIC_ANALYSIS)
             --library=posix.cfg
             --error-exitcode=1
             --verbose
-            -i src/ProvisionerTpm2.pb.cc
+            --suppress=readdirCalled
```
apldev4: What is this change for?

apldev3: Yeah so in the other implementation where we doing what `pgrep` did, it relied on a method called `readdir`, which isn't thread safe, but they deprecated the version of that was thread safe called `readdir_r` so we had to use `readdir`. CppCheck didn't appreciate that so I had to suppress the warning to get the build to pass.

The strange part is `readdir_r` was even more unsafe than the non-thread safe version, thus the deprecation, so I figured if anyone ever needed to use the function in the future they would need to use this suppression. I'm not sure if the error would go away if it was surrounded by a `mutex::lock` but the error message didn't make that suggestion at the time I was tinkering with it.

PR #56, opened 2018-11-20
[#55] Add displayTitle to Alert

### LINKED ISSUE #55: Create a display title field on Alert
The Alert class should possess a field to hold a human-readable title.

apldev1:
These changes simply add a field called 'displayTitle' to the Alert class to hold a human-readable title for each Alert instance.

Closes #55.

### DIFF
--- a/HIRS_Utils/src/main/java/hirs/data/persist/Alert.java
+++ b/HIRS_Utils/src/main/java/hirs/data/persist/Alert.java
@@ -45,6 +45,9 @@ public class Alert extends ArchivableEntity {
     @Column(name = "device_name")
     private String deviceName;
 
+    @Column(name = "display_title", nullable = true)
+    private String displayTitle;
+
     @Column(name = "details")
     private String details;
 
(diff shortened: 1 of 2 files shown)


HIRS_Utils/src/main/java/hirs/data/persist/Alert.java
```diff
@@ -45,6 +45,9 @@
     @Column(name = "device_name")
     private String deviceName;
 
+    @Column(name = "display_title", nullable = true)
```
apldev4: Do we have a reason we prefer null as a default vs. an empty string? Empty strings are less troublesome for JSPs and other code.

apldev4: only one comment

apldev3: @apldev1 any thoughts on this?

apldev1: Mostly the preference is to use null to signal that there is no content in the field, as only one alert type currently sets a value for the field.  The Portal behavior changes based on whether the field has content or not, so I figured it the code would clearly illustrate this by checking for null (alternatively, we'd still have to check for an empty string.)  I'm open to dissenting opinions, though.

apldev2: Can we get a decision on this and get this merged? I'm fine leaving it as is

apldev3: I'm fine with leaving it as `null`, people can check for it.

PR #57, opened 2018-11-26
Issue 54

iadgovuser26:
Added a ACA Users Guide and updated the install instructions in the  https://github.com/nsacyber/HIRS/tree/master/HIRS_AttestationCAPortal/src/main/webapp/docs directory which should will make it accessible in the Help page of the ACA portal. View the docs on the [issue-54 branch](https://github.com/nsacyber/HIRS/blob/issue-54/HIRS_AttestationCAPortal/src/main/webapp/docs).


apldev3:
@iadgovuser26 Just pushed an edited version of the User's Guide. At this point, I'd say I'm fine with moving forward with this. Feel free to skim through, take a look, or what have you and let me know what you think.

PR #59, opened 2018-11-28
[#58] Modify while loop condition to read all lines

### LINKED ISSUE #58: "Import Baseline" Removes PCR Entries
The "Import Baseline" (for TPM Whitelist or Blacklist) feature is removing one or more PCR entries during import. The baseline has all 24 PCR values (0-23) before performing the import.  It appears random as to which PCR(s) is/are being deleted from the TPM baseline.

mtsai1:
TPM Baselines (whitelist and blacklist) that were created via csv file upload were observed to be consistently missing exactly one record.

The cause was the while() loop condition in TPMBaselineGenerator.parseBaseline().  The BufferedReader.ready() function that governed the loop only reports on the availability of the underlying stream for reading, not the content of that stream.  By calling reader.readLine() at the end of each iteration the loop would exit after reading the last line of a file but before processing and adding it to the baseline object.  Checking reader.readLine() directly to govern the while loop avoids this problem.

Closes #58

### DIFF
--- a/HIRS_Utils/src/main/java/hirs/data/persist/TPMBaseline.java
+++ b/HIRS_Utils/src/main/java/hirs/data/persist/TPMBaseline.java
@@ -23,6 +23,7 @@ import java.util.Set;
 public abstract class TPMBaseline extends Baseline {
 
     private static final Logger LOGGER = LogManager.getLogger(TPMBaseline.class);
+    private static final String NOT_SPECIFIED = "Not Specified";
 
     @ElementCollection(fetch = FetchType.EAGER)
     @CollectionTable(name = "TPMBaselineRecords",
@@ -272,5 +273,35 @@ public abstract class TPMBaseline extends Baseline {
         }
 
         return pcrRecords.remove(record);
+    }
+
+    /**
+     * Checks the properties of FirmwareInfo, HardwareInfo, OSInfo, and TPMInfo and the contents of
+     * pcrRecords to determine if this instance of TPMBaseline is empty or not.
+     *
+     * @return true if baseline has no data
+     */
+    public boolean isEmpty() {
+        LOGGER.debug("Check for empty baseline");
+        return (firmwareInfo.getBiosReleaseDate().equals(NOT_SPECIFIED)
+                && firmwareInfo.getBiosVendor().equals(NOT_SPECIFIED)
+                && firmwareInfo.getBiosVersion().equals(NOT_SPECIFIED)
+                && hardwareInfo.getBaseboardSerialNumber().equals(NOT_SPECIFIED)
+                && hardwareInfo.getChassisSerialNumber().equals(NOT_SPECIFIED)
+                && hardwareInfo.getManufacturer().equals(NOT_SPECIFIED)
+                && hardwareInfo.getProductName().equals(NOT_SPECIFIED)
+                && hardwareInfo.getSystemSerialNumber().equals(NOT_SPECIFIED)
+                && hardwareInfo.getVersion().equals(NOT_SPECIFIED)
+                && osInfo.getDistribution().equals(NOT_SPECIFIED)
+                && osInfo.getDistributionRelease().equals(NOT_SPECIFIED)
+                && osInfo.getOSArch().equals(NOT_SPECIFIED)
+                && osInfo.getOSName().equals(NOT_SPECIFIED)
+                && osInfo.getOSVersion().equals(NOT_SPECIFIED)
+                && tpmInfo.getTPMMake().equals(NOT_SPECIFIED)
+                && tpmInfo.getTPMVersionMajor() == 0
+                && tpmInfo.getTPMVersionMinor() == 0
+                && tpmInfo.getTPMVersionRevMajor() == 0
+                && tpmInfo.getTPMVersionRevMinor() == 0
+                && pcrRecords.isEmpty());
     }
 }
--- a/HIRS_Utils/src/main/java/hirs/tpm/TPMBaselineGenerator.java
+++ b/HIRS_Utils/src/main/java/hirs/tpm/TPMBaselineGenerator.java
@@ -44,6 +44,7 @@ public class TPMBaselineGenerator {
             = LogManager.getLogger(TPMBaselineGenerator.class);
 
     private static final String KERNEL_UPDATE_BASELINE_NAME = "Kernel Update %s %s";
+    private static final String VALID_REGEX = "[0-9a-zA-Z./()_,\" -]+";
 
 
     /**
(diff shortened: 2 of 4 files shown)


mtsai1:
Fixing the error above revealed another, wherein TPMBaselineGenerator.parseBaseline() would fail to parse a file with a blank/non-null/empty last line.  This condition was interpreted as a record without a comma and would throw an exception.  Adding a regex check for the contents of each line prevents this error.

Changing the while loop condition to `dataRow = reader.readLine()` violated the checkstyle rule discouraging inner assignments.  This condition is moved into the while loop body and causes the loop to exit if a null or non-valid character value is encountered.

@iadgovuser26

HIRS_Utils/src/main/java/hirs/tpm/TPMBaselineGenerator.java
```diff
@@ -591,16 +591,15 @@ private void parseBaseline(final String baselineName, final InputStream inStream
...
         HashMap<TPMBaselineFields, String> fieldMap = new HashMap<>();
 
         try {
-            while (reader.ready()) {
-                if (StringUtils.isBlank(dataRow)) {
-                    dataRow = reader.readLine();
-                    continue;
+            while (true) {
+                dataRow = reader.readLine();
+                if (dataRow == null || !dataRow.matches("[0-9a-zA-Z./()_,\" -]+")) {
```
apldev3: Since this regex is used here and a few lines below, can we abstract them both into a `private static final String` up at the top of the class. Not quite sure what to call it. Maybe just `VALID_BASELINE_REGEX`?

HIRS_Utils/src/main/java/hirs/tpm/TPMBaselineGenerator.java
```diff
@@ -591,16 +591,15 @@ private void parseBaseline(final String baselineName, final InputStream inStream
             throws IOException, TPMBaselineGeneratorException {
         BufferedReader reader = new BufferedReader(new InputStreamReader(
                 inStream, "UTF8"));
-        String dataRow = reader.readLine();
+        String dataRow = "";
```
apldev3: Kind of a nitpick, but you don't need to initialize this with the empty String since that value gets overwritten on the first iteration of the `while` loop below. Could just be `String dataRow;`

HIRS_Utils/src/main/java/hirs/tpm/TPMBaselineGenerator.java
```diff
@@ -591,16 +591,15 @@ private void parseBaseline(final String baselineName, final InputStream inStream
...
         HashMap<TPMBaselineFields, String> fieldMap = new HashMap<>();
 
         try {
-            while (reader.ready()) {
-                if (StringUtils.isBlank(dataRow)) {
-                    dataRow = reader.readLine();
-                    continue;
+            while (true) {
```
apldev3: As part of the work on this ticket, could you do me a favor and go into the `TPMBaselineGeneratorTest` class and add an `@Test` annotation to 3 tests that are currently not running and make sure they all pass? Some of those apply to these changes, so it would be good to ensure continuing test coverage on this.

They are `generateBaselineFromCSVFileContainsInvalidRecords`, `generateBaselineFromCSVFileNullBaselineName`, and `generateBaselineFromCSVFileContainsNoRecords`.

HIRS_Utils/src/main/java/hirs/tpm/TPMBaselineGenerator.java
mtsai1: I agree that it seems guaranteed that dataRow will be overwritten in the while loop, but `String dataRow;` throws an error in the first catch block at `if (dataRow != null)` that dataRow may not have been initialized.

apldev3: Ah, good call. In that case, why not do `reader.readLine()` here, then for the while loop instead of `while(true)` do `while((dataRow != null && dataRow.matches("[0-9a-zA-Z./()_,\" -]+")`, and then move the `dataRow = reader.readLine()` that's in the `while` loop to the end of the loop instead of the beginning. Avoids this whole debacle with a small restructuring.

mtsai1: Debacle avoided.

HIRS_Utils/src/main/java/hirs/tpm/TPMBaselineGenerator.java
mtsai1: Added.  `generateBaselineFromCSVFileContainsNoRecords` did not catch an exception because of the new regex check.  I added an isEmpty() method to the TPMBaseline class so that an empty csv file that generates an empty TPMBaseline object is differentiated from a valid csv file with an empty last line.

HIRS_Utils/src/main/java/hirs/data/persist/TPMBaseline.java
```diff
@@ -273,4 +274,38 @@ public final boolean removeFromBaseline(final TPMMeasurementRecord record) {
...
     }
+
+    /**
+     * Checks the properties of FirmwareInfo, HardwareInfo, OSInfo, and TPMInfo and the contents of
+     * pcrRecords to determine if this instance of TPMBaseline is empty or not.
+     *
+     * @return true if baseline has no data
+     */
+    public boolean isEmpty() {
+    System.out.println("Check for empty baseline");
```
apldev3: Few thoughts for this. 
1. I'd stick with a Logger debug statement rather than println.
1. You could just return that giant boolean check directly rather than having the `if` statement.
1. A small Unit Test case added to `TPMBaseline` would be good to cover this new method.

HIRS_Utils/src/main/java/hirs/tpm/TPMBaselineGenerator.java
```diff
@@ -641,6 +637,10 @@ private void parseBaseline(final String baselineName, final InputStream inStream
                 TPMBaselineFields.toOSInfo(fieldMap, baseline.getOSInfo()));
             baseline.setTPMInfo(
                 TPMBaselineFields.toTPMInfo(fieldMap, baseline.getTPMInfo()));
+
+            if (baseline.isEmpty()) {
+                throw new Exception("TPM baseline is empty!");
```
apldev3: I would make this a `TpmBaselineGeneratorException` for consistency with the rest of the exception in this method, but I'm curious, were we allowed to have empty rows in CSVs before? Does this check preclude that from happening?

Like for example every baseline in the CSV is separated by an empty row.

mtsai1: The isEmpty() method was written to address the case where a csv file with valid data terminates with an empty line.  Previously, this case did not throw an error because the while loop truncated the last line of each file, empty or not.  With the modified while loop both an empty last line and a blank csv throw an error.  isEmpty() differentiates the two cases.

HIRS_Utils/src/test/java/hirs/data/persist/TPMBaselineTest.java
```diff
@@ -623,6 +623,15 @@ public final void testSetTPMInfo() {
         Assert.assertEquals(baseline.getTPMInfo(), tpmInfo);
     }
 
+    /**
+     * Test whether a baseline has data or not.
+     */
+    @Test
+    public final void testIsEmpty() {
```
apldev3: Yeah so to complement this test, can we have this be `testIsEmptyFalse`, and the a have `testIsEmptyTrue` where you just create the a TPMBaseline with its default empty constructor, and test the Assertion to verify that `IsEmpty` is true.

HIRS_Utils/src/main/java/hirs/tpm/TPMBaselineGenerator.java
```diff
@@ -641,6 +637,10 @@ private void parseBaseline(final String baselineName, final InputStream inStream
                 TPMBaselineFields.toOSInfo(fieldMap, baseline.getOSInfo()));
             baseline.setTPMInfo(
                 TPMBaselineFields.toTPMInfo(fieldMap, baseline.getTPMInfo()));
+
+            if (baseline.isEmpty()) {
+                throw new TPMBaselineGeneratorException("TPM baseline is empty!");
```
apldev3: Can we add in a test to the `TPMBaselineGeneratorTest` that has an empty CSV with an extra row, as you've suggested to make sure that this is successfully detecting the issue and throwing the expected exception here.

apldev3:
Alrighty, well I'm fine with this, thanks for sticking with me there to get the Tests and stuff knocked out. At this point, you will need to fix the Checkstyle Error in the `TPMBaselineTest` on Line 635, but aside from that you should be good.

Issue #62, opened 2018-12-10
Clean up pre-processor file expansion for error logs


iadgovuser26:
The preprocessor is expanding __FILE__ in the error message to the full path of the file where the error occurred. See Process.h. Unfortunately, it is the *full path which contains non relevant information.

apldev3:
Merge request was merged.

PR #64, opened 2018-12-11
[#62] Cleaned up preprocessor file expansion.

apldev4:
There were unnecessary references to file paths in the executable.

### DIFF
--- a/HIRS_ProvisionerTPM2/include/CommandTpm2.h
+++ b/HIRS_ProvisionerTPM2/include/CommandTpm2.h
@@ -87,7 +87,8 @@ class CommandTpm2 {
     std::string getPublicArea(const std::string& filename);
 
     std::string runTpm2CommandWithRetry(const std::string& command,
-                                        const std::string& args);
+                                            const std::string& args,
+                                            int sourceCodeLineNumber);
 
  public:
     static const char* const kDefaultIdentityClaimResponseFilename;
--- a/HIRS_ProvisionerTPM2/include/Process.h
+++ b/HIRS_ProvisionerTPM2/include/Process.h
@@ -40,18 +40,14 @@ class Process {
     std::string getOutputString() const;
 
     static std::string run(const std::string& executable,
-                           std::string sourceFileName,
-                           int sourceLineNumber,
-                           const std::string& arguments = "");
+                           const std::string& arguments,
+                           const std::string& sourceFileName,
+                           int sourceLineNumber);
 
     static bool isRunning(const std::string& executable);
 };
 
 }  // namespace utils
 }  // namespace hirs
-
-
-#define RUN_PROCESS_OR_THROW(executable, arguments)\
-    hirs::utils::Process::run(executable, __FILE__, __LINE__, arguments)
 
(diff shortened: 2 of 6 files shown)


HIRS_ProvisionerTPM2/include/CommandTpm2.h
```diff
@@ -86,8 +86,9 @@ class CommandTpm2 {
 
     std::string getPublicArea(const std::string& filename);
 
-    std::string runTpm2CommandWithRetry(const std::string& command,
-                                        const std::string& args);
+    std::string runTpm2CommandWithRetry(const std::string & command,
```
apldev3: Can we axe the spaces between the ampersand and the type of the parameter? We landed on that convention quite a while ago.

HIRS_ProvisionerTPM2/include/Process.h
```diff
@@ -39,19 +39,15 @@ class Process {
 
     std::string getOutputString() const;
 
-    static std::string run(const std::string& executable,
+    static std::string run(const std::string & executable,
```
apldev3: Same here.

HIRS_ProvisionerTPM2/src/CommandTpm2.cpp
```diff
@@ -744,12 +754,14 @@ string CommandTpm2::getPublicArea(const std::string& filename) {
     return binaryEncodedPublicArea;
 }
 
-string CommandTpm2::runTpm2CommandWithRetry(const string& command,
-                                            const string& args) {
+string CommandTpm2::runTpm2CommandWithRetry(const string & command,
```
apldev3: Same here.

HIRS_ProvisionerTPM2/src/Process.cpp
```diff
@@ -116,16 +116,16 @@ string Process::getOutputString() const {
...
  * @param sourceLineNumber line number of source file from which this method
  *                         was called
- * @param arguments the arguments including options to be passed to the
  * executable (defaults to empty string)
  */
-string Process::run(const string& executable,
-                    string sourceFileName,
-                    int sourceLineNumber,
-                    const string& arguments) {
+string Process::run(const std::string & executable,
```
apldev3: Same here about the ampersands, and additionally the `std::` qualifier is unnecessary when we're `using std::string` above.

HIRS_ProvisionerTPM2/src/CommandTpm2.cpp
```diff
@@ -150,7 +150,8 @@ void CommandTpm2::setAuthData() {
     }
 
     LOGGER.info("Attempting to set auth data.");
-    runTpm2CommandWithRetry(kTpm2ToolsTakeOwnershipCommand, argsStream.str());
+    runTpm2CommandWithRetry(kTpm2ToolsTakeOwnershipCommand, argsStream.str(),
+                            __LINE__);
```
apldev3: Seems dirty that C++ makes us do this in this way, but it is a `private` method, so that's fine I suppose.

apldev3:
Looks fine to me!

PR #70, opened 2019-01-04
[#69] Add null checks to Component Identifier Serial/Revision Trimming

### LINKED ISSUE #69: NullPointerException possible while checking component serial and revision
https://github.com/nsacyber/HIRS/blob/634d09ff5d1694028297ad6ed51722d6ba06240a/HIRS_Utils/src/main/java/hirs/validation/SupplyChainCredentialValidator.java#L500
https://github.com/nsacyber/HIRS/blob/634d09ff5d1694028297ad6ed51722d6ba06240a/HIRS_Utils/src/main/java/hirs/validation/SupplyChainCredentialValidator.java#L501

Both `component.getComponentSerial()` and `component.getComponentRevision()` can return null.  The call to `getString()` in any of the referenced lines will cause a NullPointerException on the server.  This results in a 500 error for the client.

https://github.com/nsacyber/HIRS/blob/634d09ff5d1694028297ad6ed51722d6ba06240a/HIRS_Utils/src/main/java/hirs/validation/SupplyChainCredentialValidator.java#L498
https://github.com/nsacyber/HIRS/blob/634d09ff5d1694028297ad6ed51722d6ba06240a/HIRS_Utils/src/main/java/hirs/validation/SupplyChainCredentialValidator.java#L499

Component manufacturer and model are required fields, and should not return null.  Making the assumption that the object will never be null could lead to the vague 500 error described above.

I suggest adding additional null checking for the referenced code.

apldev3:
Adds null checks to prevent the issue described in the ticket.

Closes #69

### DIFF
--- a/HIRS_Utils/src/main/java/hirs/validation/SupplyChainCredentialValidator.java
+++ b/HIRS_Utils/src/main/java/hirs/validation/SupplyChainCredentialValidator.java
@@ -491,22 +491,32 @@ public final class SupplyChainCredentialValidator implements CredentialValidator
         // 2. a list of components listed in the device info for the same manufacturer
         // Then eliminate matches from both lists. Finally, decide if the validation passes based
         // on the leftovers in the lists and the policy in place.
-
         final List<ComponentIdentifier> pcComponents = new ArrayList<>();
-        untrimmedPcComponents.forEach(component -> pcComponents.add(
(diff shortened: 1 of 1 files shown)


iadgovuser29:
Looks ok.  I tested a similar fix locally as well.

PR #73, opened 2019-01-09
#67 Add systems tests for HIRS Provisioner TPM 2.0 

busaboy1340:


.ci/system-tests/run-system-tests.sh
```diff
@@ -0,0 +1,31 @@
...
+echo "TPM2 Container ID: $tpm2_container_id"
+
+tpm2_container_status="$(docker inspect $tpm2_container_id --format='{{.State.Status}}')"
+echo "TPM2 Container Status: $tpm2_container_status"
+
+while [ $tpm2_container_status == "running" ] 
+do
+  sleep 5 
+
+  tpm2_container_status="$(docker inspect $tpm2_container_id --format='{{.State.Status}}')"
```
apldev3: For this loop we may wish to consider either throwing in an `echo` statement regarding the status of "still running". Alternatively, we could look at https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received. Although that would effectively just throw in something similar to the echo statement I mentioned above, but is even less descriptive.

My main concern here is that as we add more to these systems tests if any of them take longer than 10 minutes to run without printing out any status, then Travis will terminate the job.

.ci/system-tests/run-system-tests.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
+
+# Script to run the System Tests for HIRS
```
apldev3: If this file exists, I think we can delete the original `run-integration-tests.sh` file.

.ci/system-tests/system_test.py
```diff
@@ -0,0 +1,769 @@
+# system_test.py - implements a group of tests that run appraisals on a client and server
+
+# TODO: test_01-test_11 will need to be implemented when the flamethrower
```
apldev3: Not sure if we want to have references to the other project by name. Perhaps something to the effect of, "when the additional HIRS projects are rolled out to the new GitHub repo".

apldev3:
Looks good to me. Not sure if you'd like anyone else to look at this before we "Squash and Merge". Also, if I'm not the one to do it, please make sure this is done with the "Squash and merge" button and not the "Rebase and merge" button.

busaboy1340:
I asked one of our team members to review also. Hopefully I'll get it closed out today/tomorrow. Thanks.

PR #79, opened 2019-01-28
[#78] hirs-provisioner-tpm2 now on path after installation.

### LINKED ISSUE #78: hirs-provisioner-tpm2 not on path after installation.
hirs-provisioner-tpm2 not on path after installation. Something is going wrong in the RPM installation.

apldev4:
The post installation script in the RPM was trying to create
a symlink that already existed, so it failed before it put
hirs-provisioner-tpm2 in a directory included on the standard
path. Since the symlink should exist already anyway since
it is created by the libcurl package, we don't need the offending
line in our post installation script anyway.

Closes #78.


apldev3:
I'm a little concerned that Travis didn't kick off when you pushed to this branch, but since it's only the one line change that was expected, I'd say it's fine.

apldev3:
Everything here looks good to go. I've tested locally after the update to libcurl-devel and everything seems to be in order with regards to the Docker systems tests.

PR #84, opened 2019-01-31
[#78] hirs-provisioner-tpm2 on path after installation.

### LINKED ISSUE #78: hirs-provisioner-tpm2 not on path after installation.

apldev4:
There was a problem in the rpm-post-install.sh script
that ran as part of the CentOS7 rpm installation where
a link was being created called libcurl.so which pointed
to libcurl.so.4. If the link could not be created because
it already existed, the script would quit before finishing
and never place hirs-provisioner-tpm2 in a directory on
the PATH.

The proper solution was to link hirs-provisioner against
libcurl.so.4 so that it is clear which version of the API
was compiled against. This was not happening because
we were linking against a version of curl build by the CPR
project which was not properly embedding the SONAME in the
shared object file. By linking instead against the shared
object file distributed in the development package of
libcurl, hirs-provisioner-tpm2 now looks for libcurl.so.4
rather than the generic libcurl.so. This will prevent our
executable from breaking if libcurl.so gets updated to point
to a newer version of libcurl that uses a different API.

Closes #78.

### DIFF
--- a/HIRS_ProvisionerTPM2/CMakeLists.txt
+++ b/HIRS_ProvisionerTPM2/CMakeLists.txt
@@ -110,7 +110,8 @@ configure_file ("${CMAKE_SOURCE_DIR}/include/Tss.h.in"
 # Download necessary 3rd party libraries
 # Setup for CPR
 configure_file(lib/CPR.CMakeLists.txt.in ${CMAKE_BINARY_DIR}/lib/cpr-download/CMakeLists.txt)
-set(BUILD_CPR_TESTS OFF CACHE BOOL "" FORCE)
+set(USE_SYSTEM_CURL ON CACHE BOOL "Do not allow CPR to use its own version of curl." FORCE)
+set(BUILD_CPR_TESTS OFF CACHE BOOL "Do not waste time running CPR unit tests" FORCE)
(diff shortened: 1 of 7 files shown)


HIRS_ProvisionerTPM2/CMakeLists.txt
```diff
@@ -257,7 +258,7 @@ if (${DISTRIBUTION} STREQUAL "Ubuntu")
     set(CPACK_GENERATOR "DEB")
     set(CPACK_DEBIAN_PACKAGE_NAME "HIRSProvisionerTPM2.0")
     set(CPACK_DEBIAN_PACKAGE_SECTION "admin")
-    set(CPACK_DEBIAN_PACKAGE_DEPENDS "liblog4cplus-1.1-9(>=1.1.2), libcurlpp0(>=0.7), paccor, procps(>=3.3.0)")
+    set(CPACK_DEBIAN_PACKAGE_DEPENDS "liblog4cplus-1.1-9(>=1.1.2), libcurl4-openssl-dev(>=7.0.0), paccor, procps(>=3.3.0)")
```
apldev3: This is necessary because this package abstracts the version of `libcurl` in Ubuntu environments, i.e. this same package contains `libcurl3` or `libcurl4` depending on the Ubuntu version.

PR #86, opened 2019-01-31
[#85] Incorrect package dependencies for Ubuntu.

### LINKED ISSUE #85: Incorrect dependencies for hirs-provisioner-tpm2 RPM
`HIRS_ProvisionerTPM2/CMakeLists.txt` claims the Ubuntu deb package depends on `libcurl4-openssl-dev`; it should actually depend on `libcurl3` which provides `libcurl.so.4`. The former package depends on the later, so it would install the right package, but it is preferable to only require what is needed.

apldev4:
HIRS_ProvisionerTPM2/CMakeLists.txt claims the Ubuntu
deb package depends on libcurl4-openssl-dev; it should
actually depend on libcurl3 which provides libcurl.so.4.
The former package depends on the later, so it would
install the right package, but it is preferable to only
require what is needed.

Closes #85.

### DIFF
--- a/HIRS_ProvisionerTPM2/CMakeLists.txt
+++ b/HIRS_ProvisionerTPM2/CMakeLists.txt
@@ -258,7 +258,7 @@ if (${DISTRIBUTION} STREQUAL "Ubuntu")
     set(CPACK_GENERATOR "DEB")
     set(CPACK_DEBIAN_PACKAGE_NAME "HIRSProvisionerTPM2.0")
     set(CPACK_DEBIAN_PACKAGE_SECTION "admin")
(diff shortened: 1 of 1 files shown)


apldev3:
Based on my comment regarding this dependency in the previous MR #84, this was intentional and not "incorrect". The issue here is that `libcurl3` is only correct for some versions of Ubuntu, while `libcurl4-openssl-dev` includes `libcurl3` for some distributions and `libcurl4` for other distributions. Instead of adding in this single "incorrect" reference to `libcurl3` or actually fixing it by further complicating the Distribution Version detection logic, it's easier to leave `libcurl4-openssl-dev` and let it determine which `libcurl` is installed in the Ubuntu environment.

PR #92, opened 2019-02-12
[#91] Add Authority Information Access to Issuer field of Attribute Certificates

### LINKED ISSUE #91: Authority Information Access extension is not being displayed for Platform Attribute Certificates.
While checking new test pattern certificates for v2, it was noted that the Authority Information Access property isn't showing up in the issuer field of the certificate details page.  Upon inspection, there is no code to parse this information out for the Attribute Certificate.  CAs and EK certificates are showing this information.

cyrus-dev:
Updated code base for Attribute Certificates.  They are currently not showing Authority Information Access in the Issuer field on the certificate details page.  The code was not written to handle this or to set it.

Closes #91


HIRS_Utils/src/main/java/hirs/data/persist/certificate/Certificate.java
```diff
@@ -365,7 +365,8 @@ public Certificate(final byte[] certificateBytes) throws IOException {
                         .getInstance((DLSequence) getExtensionValue(
                                 Extension.authorityKeyIdentifier.getId()));
 
-                this.authorityInfoAccess = getAuthorityInfoAccess();
+                this.authorityInfoAccess = getAuthorityInfoAccess(x509Certificate
```
apldev3: Would it be possible to get some test coverage of at least one of these two `switch` paths (likely the X509 Certificate) and assert that this is being parsed and stored correctly?

cyrus-dev: There are parsing unit tests for both switch statements.  Are you asking for a line that checks the authority info access specifically?

apldev3: Yeah that would be good.

apldev3:
Looks good to me!

PR #93, opened 2019-02-13
[#19] General Name/DN equals functionality

### LINKED ISSUE #19: Create GeneralName Helper/Parser Class
Per @iadgovuser19 in Merge Request #16 

> Something that has a lot of room for improvement is how we deal with the GeneralName. The GeneralName can be any one of many, many types. We could/should run the getType() function to see what type of name it is before trying to parse it. It should always be an X500Name type (option '4'), because that is what the "issuer" field in an X.509v3 certificate is, which is what should be replicated into the holder of a Platform Certificate. Then, when we do the getName(), we know that the returning type can be cast as an X500Name. It is very tempting to immediately do a toString() to the X500Name, but this is not the best idea. An X500Name is a complex structure, and is subject to different interpretations when trying to read it as a string. We are using BouncyCastle when getting the string representation of the X500Name in the holder, but we are using the java.security representation when doing the same to the issuer field. Although these have always returned the same string in the past, there is no guarantee that they will always be the same. They should however be an identical ASN.1 structure, but that structure includes both SEQUENCEs (ordered sets) and SETs (not-necessarily-ordered sets). Two "identical" X500Name structures could theoretically be different sequences of bytes, with different string representations. We can either hope that this will never be different, or guarantee that we validate the similarity properly.

As a result of this, perhaps we should make some type of small Helper class in `HIRS_Utils` that can be employed around the rest of the HIRS subprojects as needed.

cyrus-dev:
The code was currently using String.equals() on different DNs and this would fail when technically the contents were the same.  Any difference, capitals, RDNs in a different order or different RDN separators, would cause this to fail.  Instead of a new class, bouncy castle's X500Name class' equals handles this properly.  An already in use class was updated to include a method that compares two values.  An addition to another class was only done because there have been certificates that used '+' sign as a separator and this causes X500Name to fail as well.  So the check for non-comma separators was added.  This should be updated on newer types of separators being used.

Closes #19


HIRS_Utils/src/main/java/hirs/utils/Functional.java
```diff
@@ -38,4 +43,25 @@ private Functional() {
...
+     * This method can be used to compare the distinguished names given from
+     * certificates.  This compare uses X500Name class in bouncy castle, which
+     * compares the RDNs and not the string itself.  The method will check
+     * for '+' and replace them, X500Name doesn't do this.
+     *
+     * @param nameValue1 first general name to be used
+     * @param nameValue2 second general name to be used
+     * @return true if the values match based on the RDNs, false if not
+     */
+    public static boolean x500NameCompare(final String nameValue1, final String nameValue2) {
```
apldev3: This should go into a new Utility class like `X500Utils` or `BouncyCastleUtils`. Based on the only other method in here this class looks to have been designed to support methods often natively supported by functional programming languages, such as select, take, map, reduce, etc.

HIRS_AttestationCAPortal/src/main/java/hirs/attestationca/portal/util/CertificateStringMapBuilder.java
```diff
@@ -140,12 +141,13 @@ public static Certificate containsAllChain(
...
         }
 
-        for (Certificate issuerCert: issuerCertificates) {
+        for (Certificate issuerCert : issuerCertificates) {
             try {
                 // Find the certificate that actually signed this cert
                 if (certificate.isIssuer(issuerCert)) {
                     //Check if it's root certificate
-                    if (issuerCert.getIssuer().equals(issuerCert.getSubject())) {
+                    if (Functional.x500NameCompare(issuerCert.getIssuer(),
```
apldev3: This should shift to whatever the new `Utils` class is.

HIRS_Utils/src/main/java/hirs/utils/Functional.java
apldev3: Additionally, whatever class this goes into should have a new test class written to test this comparison. We should have the test certificates to support a Unit Test of this.

apldev3:
This is also looking good to me!

PR #94, opened 2019-02-13
[#72] Supply Chain Validator fix and update

### LINKED ISSUE #72: PlatformSerial optional field in Subject Alternative Name
PlatformSerial is defined as an optional value in the Subject Alternative Name.  Most platforms should have the information that goes into this value available.  If they don't, then the platform certificate might not include the value.  

Field Name | RFC 5280 Type | Value | Field Status
------------ | ------------ | ------------ | ------------ 
| Subject Alternative Name | GeneralName<br/>directoryName | PlatformManufacturerStr<br/>PlatformModel<br/>PlatformVersion<br/>PlatformSerial (optional)<br/>PlatformManufacturerId (optional) | MUST<br/>critical/non-critical<br/>(depending on subject) |

The specification says:
> During certificate validation, the Privacy-CA MUST check that the platform manufacturer, model, version, serial numbers, and manufacturer ID are acceptable.

Currently, if the ACA finds that the value is not present in the certificate, then it throw a null error.

Since the value is optional, the appraisal should fail only if the value is not the same in the device info and the certificate.  If neither have a platform serial number, appraisal should continue.

cyrus-dev:
This fix correct an IllegalStateException for the SupplyChainValidator when all policy settings are true.  When trying to remove a value from the iterator in the validator, the item was null and caused this issue.  This also takes out the Platform Serial as a required field.

Closes #72


HIRS_Utils/src/main/java/hirs/validation/SupplyChainCredentialValidator.java
```diff
@@ -678,6 +688,26 @@ private static boolean requiredPlatformCredentialFieldIsNonEmptyAndMatches(
             return false;
         }
 
+        return platformCredentialFieldMatches(platformCredentialFieldName,
+                platformCredentialFieldValue, otherValue);
+    }
+
+    private static boolean notRequiredPlatformCredentialFieldMatchesNotNullAndMatches(
```
apldev3: I would say that this method should get Javadoc similar to the one it mirrors above for consistency. Also, I would consider renaming this to something that's a bit easier to understand `notRequired` could be `optional`.

Additionally, the name says it matches not null, but it looks to match explicitly on null, so perhaps the name should be `optionalPlatformCredentialFieldNullOrMatches`.

HIRS_Utils/src/main/java/hirs/validation/SupplyChainCredentialValidator.java
```diff
@@ -493,8 +495,8 @@ private static boolean validateV2p0PlatformCredentialComponentsExpectingExactMat
         // on the leftovers in the lists and the policy in place.
         final List<ComponentIdentifier> pcComponents = new ArrayList<>();
         for (ComponentIdentifier component : untrimmedPcComponents) {
-            DERUTF8String componentSerial = null;
-            DERUTF8String componentRevision = null;
+            DERUTF8String componentSerial = new DERUTF8String("");
+            DERUTF8String componentRevision = new DERUTF8String("");
```
apldev3: Can we update the `SupplyChainValidatorTest` in some way to indicate that the bug is actually fixed.

apldev3:
Looks good to me!

PR #97, opened 2019-02-21
Modify build scripts to enable installation on Amazon Linux

### LINKED ISSUE #90: Need support for Amazon Linux
Amazon AWS has a set of supported Amazon Machine Images (AMI). "Amazon Linux" is a commonly used AMI. Amazon Linux defines the OS version within the /etc/system-release file which ,as example, currently contains "Amazon Linux release 2 (Karoo)" for the Amazon Linux  AMI. 

Currently the HIRS RPMs are  looking for similar files (e.g. /etc/rehat-release) but will not install properly on Amazon Linux. Adding a /etc/redhat-release file in the AWS instance will fix the problem for that install. To support Amazon linux a check for the /etc/system-release file should be put in the install scripts if /etc/redhat is not found.

chubtub:
Modified the build scripts to parse the Amazon Linux version as appropriate.  Default behavior is to treat version 2 the same as RHEL7.

Closes #90

### DIFF
--- a/.ci/docker/Dockerfile.amazonlinux2
+++ b/.ci/docker/Dockerfile.amazonlinux2
@@ -1,0 +1,17 @@
+FROM amazonlinux:2
+
+MAINTAINER apl.dev3@jhuapl.edu
+
+# Install packages for building HIRS
+RUN yum -y update && yum clean all
+RUN yum groupinstall -y "Development Tools"
+RUN yum install -y wget java-1.8.0-openjdk-devel protobuf-compiler rpm-build cmake make git gcc-c++ doxygen graphviz python libssh2-devel openssl protobuf-devel tpm2-tss-devel trousers-devel libcurl-devel
+
+# Install EPEL
+WORKDIR /tmp
+RUN wget -O epel.rpm -nv https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
+RUN yum install -y ./epel.rpm
+RUN yum install -y cppcheck log4cplus-devel re2-devel
+
+# Set Environment Variables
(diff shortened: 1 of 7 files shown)


apldev3:
This looks good to me. Has this been tested on an AWS image?

apldev3:
Hey @mtsai1 just a headsup that I created a Docker Image to test the packaging on Amazon Linux and it's not working properly. It does fine right up until it needs to build HIRS_ProvisionerTPM2 and then it gets a little wonky.

Best I can reason is that `make` is just a little bit different on the Amazon Linux container compared to basic CentOS or RHEL. When I get the container to just do a CMake I can go in and `make` everything it generated just fine, but if I try to do `make` from within the Amazon Linux container it appears to not be linking the libraries properly. Will need to do additional exploration of this issue to figure out the difference and account for it in the `CMakeLists.txt`.

mtsai1:
This issue was raised to address the centos packaging script printing an error message because it could not detect the centos version.  The changes resolve packaging errors and generate rpms that are successfully installed on Amazon Linux.  The cmake errors should be addressed in a separate issue.

apldev3:
Alright, that's fair, we can set that up.

apldev3:
@mtsai1 I went ahead and commented out that build step, but left it in with a TODO that can be addressed in a separate issue.

apldev3:
Issue #102 created to track the building/packaging issue on Amazon Linux. For the mean time the installation of the RPM should work per this ticket.

apldev3:
I'll leave it to @iadgovuser26 to verify that this works for their needs.

apldev3:
@iadgovuser26 same question as in the other ticket. Are we good here?

PR #100, opened 2019-02-25
Updates to allow for TPM 2.0 quote.

### LINKED ISSUE #99: Updates for TPM 2.0 Quote
Updates required to support TPM 2.0 Quote include:

- [ ] changes to TPM Report to hold TPM 2 Quote value.

apldev4:
XML for TPM report can not hold raw TPM 2.0 quote.

Closes #99.

### DIFF
--- a/HIRS_Provisioner/scripts/install/hirs-provisioner.sh
+++ b/HIRS_Provisioner/scripts/install/hirs-provisioner.sh
@@ -9,6 +9,7 @@ CERTS_DIR="/etc/hirs/provisioner/certs/"
 HIRS_SITE_CONFIG="/etc/hirs/hirs-site.config"
 HIRS_PROVISIONER_CONFIG="/etc/hirs/provisioner/hirs-provisioner-config.sh"
 HIRS_PROVISIONER_SCRIPT="/usr/share/hirs/provisioner/bin/HIRS_Provisioner"
+HIRS_PROVISIONER_2_0_SCRIPT="/usr/local/bin/hirs-provisioner-tpm2"
 HIRS_PROVISIONER_PROPERTIES="/etc/hirs/provisioner/provisioner.properties"
 
 if [ "$EUID" != "0" ]; then
@@ -105,10 +106,16 @@ function CheckProvisionPrereqsAndDoProvisioning {
 
 function Provision {
     # Provisioner will only retain one {uuid}.cer credential; remove any existing *.cer files.
-    echo "----> Removing old provisioner scripts, if any"
-    rm -f $CERTS_DIR/*.cer
+    echo "----> Removing old attestation credentials, if any"
+    rm -f $CERTS_DIR/*.cer /etc/hirs/ak.cer
     echo "----> Provisioning TPM"
-    $HIRS_PROVISIONER_SCRIPT $CLIENT_HOSTNAME || { echo "----> Failed to provision TPM"; exit 1; }
+
+    if [ -f $HIRS_PROVISIONER_2_0_SCRIPT ]
+    then
+        $HIRS_PROVISIONER_2_0_SCRIPT provision || { echo "----> Failed to provision TPM 2.0"; exit 1; }
+    else
+        $HIRS_PROVISIONER_SCRIPT $CLIENT_HOSTNAME || { echo "----> Failed to provision TPM"; exit 1; }
+    fi
 }
 
 
--- a/HIRS_ProvisionerTPM2/CMakeLists.txt
+++ b/HIRS_ProvisionerTPM2/CMakeLists.txt
@@ -140,11 +140,11 @@ list(APPEND REQUIRED_LIBS ${PROTOBUF_LIBRARY})
 # Define the TPM 2.0 Provisioner Library
 add_subdirectory(src)
 
-# Set Project Header/Source files
-set(EXECUTABLE_SOURCE_FILES src/TPM2_Provisioner.cpp)
-
 # Create project executable
-add_executable(${PROJECT_NAME} ${EXECUTABLE_SOURCE_FILES} ${PROJECT_CONFIG_FILES})
+add_executable(${PROJECT_NAME} src/TPM2_Provisioner.cpp ${PROJECT_CONFIG_FILES})
+
+# In TPM 2.0 land, there is currently not a way to fetch the TPM version info
+add_executable(tpm_version src/tpm_version.cpp)
 
 # Link necessary libraries
 target_link_libraries(${PROJECT_NAME} TPM2_PROVISIONER_LIBRARY)
@@ -159,13 +159,25 @@ if(CMAKE_THREAD_LIBS_INIT)
 endif()
 
 # Set commands for installation of project on target system (i.e. "make install")
-install(TARGETS ${PROJECT_NAME}
+install(TARGETS ${PROJECT_NAME} tpm_version
         DESTINATION "bin")
 install(FILES config/log4cplus_config.ini DESTINATION /etc/hirs/TPM2_Provisioner)
 install(FILES scripts/tpm_aca_provision DESTINATION /usr/local/bin PERMISSIONS OWNER_READ OWNER_WRITE OWNER_EXECUTE GROUP_READ GROUP_EXECUTE WORLD_READ WORLD_EXECUTE)
 install(DIRECTORY DESTINATION ${/var/log/hirs/provisioner}
         DIRECTORY_PERMISSIONS OWNER_READ OWNER_WRITE OWNER_EXECUTE GROUP_READ
                               GROUP_EXECUTE WORLD_READ WORLD_EXECUTE)
+
+install(DIRECTORY DESTINATION ${/etc/hirs/provisioner}
+        DIRECTORY_PERMISSIONS OWNER_READ OWNER_WRITE OWNER_EXECUTE GROUP_READ
+                              GROUP_EXECUTE WORLD_READ WORLD_EXECUTE)
+install(FILES ../HIRS_Provisioner/src/main/resources/defaults.properties DESTINATION /etc/hirs/provisioner RENAME provisioner.properties)
+install(FILES ../HIRS_Provisioner/hirs-provisioner-config.sh DESTINATION /etc/hirs/provisioner
+        PERMISSIONS OWNER_READ OWNER_WRITE OWNER_EXECUTE GROUP_READ
+                    GROUP_EXECUTE WORLD_READ WORLD_EXECUTE)
+install(FILES ../HIRS_Provisioner/scripts/install/hirs-provisioner.sh DESTINATION /etc/hirs/provisioner
+        PERMISSIONS OWNER_READ OWNER_WRITE OWNER_EXECUTE GROUP_READ
+                    GROUP_EXECUTE WORLD_READ WORLD_EXECUTE)
+
 execute_process(COMMAND cp ../HIRS_Utils/src/main/resources/logging.properties ./config/
         RESULT_VARIABLE result
         WORKING_DIRECTORY ${CMAKE_SOURCE_DIR})
--- a/HIRS_ProvisionerTPM2/package/postinst
+++ b/HIRS_ProvisionerTPM2/package/postinst
@@ -1,8 +1,10 @@
 HIRS_SITE_CONFIG="/etc/hirs/hirs-site.config"
 
 mkdir -p /var/log/hirs/provisioner
+mkdir -p /etc/hirs/provisioner/certs
 ln -s -f /usr/local/bin/hirs-provisioner-tpm2 /usr/sbin/hirs-provisioner-tpm2
 ln -s -f /usr/local/bin/tpm_aca_provision /usr/sbin/tpm_aca_provision
+ln -s -f /usr/local/bin/tpm_version /usr/sbin/tpm_version
 
 if [ ! -f $HIRS_SITE_CONFIG ]; then
     # Create template site config if it does not exist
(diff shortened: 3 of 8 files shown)


HIRS_Utils/src/main/java/hirs/data/persist/tpm/PcrInfoShort.java
```diff
@@ -199,17 +203,33 @@ public final int hashCode() {
...
+
+    /**
+     * Calculates the digest of the PCR values the same way a TPM computes the digest contained in
+     * the quote. Useful for TPM appraisal and for ensuring the digest of the collected PCR values
+     * match the digest in the quote.
+     *
+     * @param messageDigest message digest algorithm to use
+     * @return byte array containing the digest
+     */
+    public final byte[] getCalculatedDigestTpmV1p2(final MessageDigest messageDigest) {
```
apldev3: This seems like this could be a `private` method since the TPM 1.2 should only ever pass in "SHA-1" as a MessageDigest.

HIRS_Utils/src/main/java/hirs/data/persist/tpm/PcrInfoShort.java
```diff
@@ -235,6 +255,32 @@ public final int hashCode() {
...
 
+    /**
+     * Calculates the digest of the PCR values the same way a TPM computes the digest contained in
+     * the quote. Useful for TPM appraisal and for ensuring the digest of the collected PCR values
+     * match the digest in the quote.
+     *
+     * @param messageDigest message digest algorithm to use
+     * @return byte array containing the digest
+     */
+    public final byte[] getCalculatedDigestTpmV2p0(final MessageDigest messageDigest) {
```
apldev3: This could be a `private` method as we could force the user to go through the more general `getCalculatedDigest` method. I could also see leaving it `public` to let a user choose the algorithm for TPM 2.0. That being said, you could just overload the `getCalculatedDigest` and write another version of that that accepts an argument for MessageDigest which it can then pass through to this as a `private` method.

HIRS_ProvisionerTPM2/src/tpm_version.cpp
```diff
@@ -0,0 +1,11 @@
+#include <iostream>
+
+using std::cout;
+using std::endl;
+
+int main(void) {
+    cout << "Chip Version: 2.0.0.0" << endl;
+    cout << "TPM Vendor ID: UNKN" << endl;
```
apldev3: There's really no official way to get the version and vendor from the chip? I could have sworn that we had something like that way back at the start of the TPM 2.0 Provisioner project that Andrew had implemented.

HIRS_ProvisionerTPM2/package/postinst
```diff
@@ -1,6 +1,7 @@
 HIRS_SITE_CONFIG="/etc/hirs/hirs-site.config"
 
 mkdir -p /var/log/hirs/provisioner
+mkdir -p /etc/hirs/provisioner/certs
 ln -s -f /usr/local/bin/hirs-provisioner-tpm2 /usr/sbin/hirs-provisioner-tpm2
 ln -s -f /usr/local/bin/tpm_aca_provision /usr/sbin/tpm_aca_provision
```
apldev3: Wouldn't you need the `ln -s -f /usr/local/bin/tpm_version /usr/sbin/tpm_version` for this Ubuntu package as well?

HIRS_ProvisionerTPM2/CMakeLists.txt
```diff
@@ -140,11 +140,11 @@ list(APPEND REQUIRED_LIBS ${PROTOBUF_LIBRARY})
...
 
-# Set Project Header/Source files
-set(EXECUTABLE_SOURCE_FILES src/TPM2_Provisioner.cpp)
-
 # Create project executable
-add_executable(${PROJECT_NAME} ${EXECUTABLE_SOURCE_FILES} ${PROJECT_CONFIG_FILES})
+add_executable(${PROJECT_NAME} src/TPM2_Provisioner.cpp ${PROJECT_CONFIG_FILES})
+
+# In TPM 2.0 land, there is currently not a way to fetch the TPM version info
+add_executable(tpm_version src/tpm_version.cpp)
```
apldev3: See comment in `tpm_version.cpp`.

HIRS_ProvisionerTPM2/src/tpm_version.cpp
apldev3: I just looked and yeah there is a `tpm2_getcap` tool that could be used to get the info. See the man page, [here](https://github.com/tpm2-software/tpm2-tools/blob/master/man/tpm2_getcap.1.md). That being said it could get tricky as I'm fairly certain that older versions of that tool went under a different name. Likely `tpm2_getcapabilities` or something. That being said, if we wanted to go this route, I'd recommend adding another command to the provisioner to get this information rather than a separate utility binary, like this, since we have a lot of the infrastructure in place for interfacing directly with the TPM there, of course.

HIRS_Provisioner/scripts/install/hirs-provisioner.sh
```diff
@@ -9,6 +9,7 @@ CERTS_DIR="/etc/hirs/provisioner/certs/"
 HIRS_SITE_CONFIG="/etc/hirs/hirs-site.config"
 HIRS_PROVISIONER_CONFIG="/etc/hirs/provisioner/hirs-provisioner-config.sh"
 HIRS_PROVISIONER_SCRIPT="/usr/share/hirs/provisioner/bin/HIRS_Provisioner"
+HIRS_PROVISIONER_2_0_SCRIPT="/usr/local/bin/hirs-provisioner-tpm2"
```
apldev3: What was the impetus for adding all of the TPM 2.0 stuff to the older provisioner's script?

HIRS_ProvisionerTPM2/CMakeLists.txt
```diff
@@ -159,13 +159,25 @@ if(CMAKE_THREAD_LIBS_INIT)
...
 install(FILES scripts/tpm_aca_provision DESTINATION /usr/local/bin PERMISSIONS OWNER_READ OWNER_WRITE OWNER_EXECUTE GROUP_READ GROUP_EXECUTE WORLD_READ WORLD_EXECUTE)
 install(DIRECTORY DESTINATION ${/var/log/hirs/provisioner}
         DIRECTORY_PERMISSIONS OWNER_READ OWNER_WRITE OWNER_EXECUTE GROUP_READ
                               GROUP_EXECUTE WORLD_READ WORLD_EXECUTE)
+
+install(DIRECTORY DESTINATION ${/etc/hirs/provisioner}
+        DIRECTORY_PERMISSIONS OWNER_READ OWNER_WRITE OWNER_EXECUTE GROUP_READ
+                              GROUP_EXECUTE WORLD_READ WORLD_EXECUTE)
+install(FILES ../HIRS_Provisioner/src/main/resources/defaults.properties DESTINATION /etc/hirs/provisioner RENAME provisioner.properties)
+install(FILES ../HIRS_Provisioner/hirs-provisioner-config.sh DESTINATION /etc/hirs/provisioner
```
apldev3: I'm guessing this has to do with the other projects that aren't touched on in this repo?

HIRS_Utils/src/main/java/hirs/data/persist/tpm/PcrInfoShort.java
```diff
@@ -250,4 +296,18 @@ public final int hashCode() {
...
+        List<TPMMeasurementRecord> pcrs = this.getPcrComposite().getPcrValueList();
+        if (pcrs.size() == 0) {
+            // TODO figure out a reasonable case
+            // it's the case of an empty pcrmask, so maybe it doesn't matter
+            return false;
+        }
+
+        Digest hash = pcrs.get(0).getHash();
+        // check if the hash algorithm is SHA 1, if so it's TPM 1.2, if not it's TPM 2.0
+        return hash.getAlgorithm() == DigestAlgorithm.SHA1;
```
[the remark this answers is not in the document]
apldev2: So, TPM 1.2 supports SHA1 and TPM 2.0 supports SHA 1, 256, and more (especially due to algorithm agility). For our purposes, we hard coding TPM 2.0 to only report SHA 256. Because we are reusing the 1.2 data structures for appraisal, this is the simplest way of making this determination without adding additional data to the class, which was a bit of a feature requirement.

HIRS_Utils/src/main/java/hirs/data/persist/tpm/PcrInfoShort.java
apldev2: I need to double check this, but I'm not sure that's actually possible. Several of the methods in this class are actually needed by the Appraiser to do the appraisal. Their use isn't restricted simply to within this class.

apldev2: Ah ok, I see your point. This is only referenced via getCalculatedDigest.

HIRS_ProvisionerTPM2/CMakeLists.txt
apldev2: I believe the client work requires the provisioner.properties config. This is just updating the 2.0 provisioner to drop that file on the machine because it is currently only installed by the 1.2 provisioner.

HIRS_ProvisionerTPM2/src/tpm_version.cpp
apldev2: This was intended only to be a stopgap. We were planning on addressing this in a follow-on issue. In the interest of getting this through, I'd probably be in favor of leaving this as is and addressing it in the subsequent work.

HIRS_Utils/src/main/java/hirs/data/persist/tpm/PcrInfoShort.java
```diff
@@ -250,4 +296,18 @@ public final int hashCode() {
 
         return byteBuffer.array();
     }
+
+    private boolean isTpm1() {
+        // need to get an individual PCR and measure length to determine SHA1 v SHA 256
+        List<TPMMeasurementRecord> pcrs = this.getPcrComposite().getPcrValueList();
+        if (pcrs.size() == 0) {
+            // TODO figure out a reasonable case
```
apldev2: TODO. I think I may have actually written this line. I think what I suggest in the following line is reasonable.

HIRS_Provisioner/scripts/install/hirs-provisioner.sh
apldev2: It looks like this script is still installed on the TPM 2.0 client. I think it's used to standardize the provisioning interface.

HIRS_Utils/src/main/java/hirs/data/persist/tpm/QuoteSignature.java
```diff
@@ -76,6 +81,29 @@ protected QuoteSignature() {
      *            SignatureValue that contains the Quote2 signature, may be null
      */
     public QuoteSignature(final byte[] signatureValue) {
```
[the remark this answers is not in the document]
apldev2: fixed

HIRS_Utils/src/main/java/hirs/data/persist/tpm/PcrInfoShort.java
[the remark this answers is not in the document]
apldev2: going to mark this resolved

HIRS_Utils/src/main/java/hirs/data/persist/tpm/PcrInfoShort.java
apldev2: fixed

HIRS_ProvisionerTPM2/src/tpm_version.cpp
apldev2: tabling this for a new issue

HIRS_ProvisionerTPM2/CMakeLists.txt
apldev2: marking resolved

HIRS_ProvisionerTPM2/CMakeLists.txt
apldev2: marking resolved

HIRS_Utils/src/main/java/hirs/data/persist/tpm/PcrInfoShort.java
apldev2: fixed

HIRS_ProvisionerTPM2/package/postinst
apldev2: I didn't have a definitive answer on this, so I just went ahead and added it.

HIRS_Utils/src/main/java/hirs/data/persist/tpm/PcrInfoShort.java
apldev2: fixed

HIRS_Provisioner/scripts/install/hirs-provisioner.sh
```diff
@@ -105,10 +106,16 @@ function CheckProvisionPrereqsAndDoProvisioning {
...
-    echo "----> Removing old provisioner scripts, if any"
-    rm -f $CERTS_DIR/*.cer
+    echo "----> Removing old attestation credentials, if any"
+    rm -f $CERTS_DIR/*.cer /etc/hirs/ak.cer
     echo "----> Provisioning TPM"
-    $HIRS_PROVISIONER_SCRIPT $CLIENT_HOSTNAME || { echo "----> Failed to provision TPM"; exit 1; }
+
+    if [ -f $HIRS_PROVISIONER_2_0_SCRIPT ]
+    then
+        $HIRS_PROVISIONER_2_0_SCRIPT provision || { echo "----> Failed to provision 2.0 TPM"; exit 1; }
```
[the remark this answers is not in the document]
apldev2: fixed

apldev2:
Ok, I think I addressed all the comments if you're satisfied @apldev3

apldev3:
Yeah it looks good to me. Just need to write up an issue about the TPM Version Detection.

PR #101, opened 2019-02-28
[#95] Resolve Log4cplus Error complaining about no appenders

### LINKED ISSUE #95: log4cplus error printed during provisioning.
When provisioning a tpm2 log4cplus prints an initialization error 
```
TPM 2.0 detected.
log4cplus:ERROR No appenders could be found for logger (root).
log4cplus:ERROR Please initialize the log4cplus system properly.--> Provisioning
----> Collecting endorsement credential from TPM
----> Creating attestation key
----> Collecting platform credential from TPM
...
```

apldev3:
The main issue that was causing the error to appear was the fact that we statically create the Loggers for 5 of our classes in `HIRS_ProvisionerTPM2`. The way the `logging.properties` loading method was written in `Logger.cpp` it attempted to Log the new log level of the loggers.

The problem was that this all happens before we even get to the `main` method and initialize `Log4cplus` to begin with. Since Log4cplus wasn't initialized, when it attempted to update the log level for these `static` loggers and then Log what their new log levels were, it would throw the error.

This pull request just removes those unnecessary log statements and the problem is resolved.

Closes #95

### DIFF
--- a/HIRS_ProvisionerTPM2/include/Logger.h
+++ b/HIRS_ProvisionerTPM2/include/Logger.h
@@ -18,7 +18,7 @@ namespace log {
 class Logger {
  private:
     static const char* const kDefaultProvisionerLoggerName;
-    static const char* const PROP_FILE_LOC;
+    static const char* const kPropFileLocation;
 
     const log4cplus::Logger kLogger;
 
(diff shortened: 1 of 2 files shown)


apldev3:
As a follow-up all of this was determined via `strace` and observing when the `logging.properties` file was checked vs when the `main` function actually started.

`Logging.properties` was checked 6 times before we get to the `main` function. Once for the `root` logger, and then 5 times for each of the classes where we statically create a logger.

apldev3:
@iadgovuser26 is this good to merge? Did you get to test it on your end?

apldev3:
If it's good enough for @apldev4 it's good enough for me, merging.

PR #104, opened 2019-03-08
[#103] Blanks changed to 'Not Specified'

### LINKED ISSUE #103: Blank String Provision Sync
This issue is related [#96], when doing some testing afterwards on unrelated issues there were several problems that occurred.

1. If enough components mismatched occur, the message length that is saved to the DB is too long for the default.  
2. The components that have been coming up mismatched are mismatched because of blank entries not lining up.  Either paccor has a components that is "Not Specified", sometimes the component matched is a null, sometimes a component is blank.  

This issue will be to sync all components in ComponentInfo and ComponentIdentifier to "Not Specified" when normally set to null or "" or not set at all if the provided information is missing from paccor.

cyrus-dev:
The suppy chain validation has reported non-matches for components that are initialized as null or blank or never initialized by code but by COTS. The end results is validation failure because the certificate compare will be null and the device report will list blank or Not Specified. This issue is to synch all to 'Not Specified'. Previous blank was set but that still caused issue with a blank compared to a 'Not Specified'.

Closes #103

### DIFF
--- a/HIRS_Utils/src/main/java/hirs/data/persist/ComponentInfo.java
+++ b/HIRS_Utils/src/main/java/hirs/data/persist/ComponentInfo.java
@@ -194,9 +194,13 @@ public class ComponentInfo implements Serializable {
         this.componentModel = componentModel.trim();
         if (componentSerial != null) {
             this.componentSerial = componentSerial.trim();
+        } else {
+            this.componentSerial = DeviceInfoReport.NOT_SPECIFIED;
         }
         if (componentRevision != null) {
             this.componentRevision = componentRevision.trim();
+        } else {
+            this.componentRevision = DeviceInfoReport.NOT_SPECIFIED;
         }
     }
 
@@ -216,9 +220,8 @@ public class ComponentInfo implements Serializable {
                                      final String componentModel,
                                      final String componentSerial,
                                      final String componentRevision) {
-        return !(
-                StringUtils.isEmpty(componentManufacturer)  || StringUtils.isEmpty(componentModel)
-        );
+        return !(StringUtils.isEmpty(componentManufacturer)
+                || StringUtils.isEmpty(componentModel));
     }
 
     @Override
--- a/HIRS_Utils/src/main/java/hirs/data/persist/SupplyChainValidation.java
+++ b/HIRS_Utils/src/main/java/hirs/data/persist/SupplyChainValidation.java
@@ -48,7 +48,7 @@ public class SupplyChainValidation extends ArchivableEntity {
             joinColumns = { @JoinColumn(name = "validation_id", nullable = false) })
     private final List<Certificate> certificatesUsed;
 
-    @Column
+    @Column(columnDefinition = "VARCHAR(700)")
     private final String message;
 
     /**
--- a/HIRS_Utils/src/main/java/hirs/data/persist/certificate/attributes/ComponentIdentifier.java
+++ b/HIRS_Utils/src/main/java/hirs/data/persist/certificate/attributes/ComponentIdentifier.java
@@ -1,5 +1,6 @@
 package hirs.data.persist.certificate.attributes;
 
+import hirs.data.persist.DeviceInfoReport;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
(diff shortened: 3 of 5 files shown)


HIRS_Utils/src/main/java/hirs/data/persist/ComponentInfo.java
```diff
@@ -216,9 +220,8 @@ public static boolean isComplete(final String componentManufacturer,
                                      final String componentModel,
                                      final String componentSerial,
                                      final String componentRevision) {
-        return !(
-                StringUtils.isEmpty(componentManufacturer)  || StringUtils.isEmpty(componentModel)
-        );
+        return !(StringUtils.isEmpty(componentManufacturer)
+                || StringUtils.isEmpty(componentModel));
```
apldev3: This is more of a stylistic issues, but I know I reason about this better when we do a simplification of these lines to simply `return !StringUtils.isEmpty(componentManufacturer) && !StringUtils.isEmpty(componentModel);` This keeps the statement to "componentManufacturer is not empty AND componentModel is not empty". It will work either way, but one seems easier to interpret at a passing glance.

HIRS_Utils/src/main/java/hirs/data/persist/SupplyChainValidation.java
```diff
@@ -48,7 +48,7 @@
             joinColumns = { @JoinColumn(name = "validation_id", nullable = false) })
     private final List<Certificate> certificatesUsed;
 
-    @Column
+    @Column(columnDefinition = "VARCHAR(700)")
```
apldev3: How did we land on 700 as the size here?

HIRS_Utils/src/main/java/hirs/data/persist/certificate/attributes/ComponentIdentifier.java
```diff
@@ -55,11 +56,11 @@
...
-        componentManufacturer = null;
-        componentModel = null;
-        componentSerial = null;
-        componentRevision = null;
+        componentClass = DeviceInfoReport.NOT_SPECIFIED;
+        componentManufacturer = new DERUTF8String(DeviceInfoReport.NOT_SPECIFIED);
+        componentModel = new DERUTF8String(DeviceInfoReport.NOT_SPECIFIED);
+        componentSerial = new DERUTF8String(DeviceInfoReport.NOT_SPECIFIED);
+        componentRevision = new DERUTF8String(DeviceInfoReport.NOT_SPECIFIED);
         componentManufacturerId = null;
```
apldev3: I wonder if this `componentManufacturerId` should be `componentManufacturerOid` since that's what the type is. Makes it more immediately clear that this isn't a `String` or `Guid` or some kind.

apldev3:
Definitely in favor of moving to some type of consistent value, so this is a good MR overall!

HIRS_Utils/src/main/java/hirs/data/persist/ComponentInfo.java
cyrus-dev: Will look over again.  Looking at this again, it looks odd to me as well.

HIRS_Utils/src/main/java/hirs/data/persist/SupplyChainValidation.java
cyrus-dev: When testing with the added fields, if they all didn't match the server failed because the default sql varchar size wasn't large enough to store.  I picked a large number since the contents of those components can be of any given length based on the manufacturer.

HIRS_Utils/src/main/java/hirs/data/persist/certificate/attributes/ComponentIdentifier.java
cyrus-dev: I'm not sure what you mean.  At the moment, we don't have an example of what this could end up being.  I hope it does come out as a String.  Don't have a test pattern yet.  The java type is ANS1ObjectIdentifier but it isn't an OID, as in 2.23.133.18.3.1.

It is PrivateEnterpriseNumber, and I don't know what that'll be, so I made it a default value.  It may actually be better to set it to String.

cyrus-dev:
As an update for this request, I will be revisiting this after talking with the paccor creator.  As stated above this was a push to add changes I made while resolving an issue with the ACA validating a provisioned emulator.  

However since 'Not Specified' is also a validate entry therefore I don't want to populate absent fields with that value.  Once I'm done with #107  I will revisit this change and re-push.

HIRS_Utils/src/main/java/hirs/data/persist/SupplyChainValidation.java
apldev3: Alrighty sounds good to me.

HIRS_Utils/src/main/java/hirs/data/persist/certificate/attributes/ComponentIdentifier.java
apldev3: Could be, I'll leave that call to you, as I'm not familiar with PrivateEnterpriseNumbers.

PR #113, opened 2019-03-13
[#24] Implementation of Component Class field

### LINKED ISSUE #24: Implement Support for the Component Class field within the  ComponentIdentifier .
New test pattens are introducing a new field within the ComponentIdentifier :
OCTET STRING SIZE(4). This needs to be accounted for when parsing the Platform Credential. No processing of this field is current needed.

cyrus-dev:
This is new code that parses a new field in the upcoming TCG spec for  the platform components fields. The new field indicates the type of hardware (ex Memory - DDR3). This information wasn't provided before so it wasn't always clear what the component was. The new information is provided in a json file. A unit test was created to test the different variations. This commit does not include hooks in the base code to use this class yet. This commit is mainly to include the added library and correct bug and checkstyle issues associated with the new code.

Closes #24


cyrus-dev:
Redone on #114

PR #114, opened 2019-03-13
[#24] Implementation of Component Class field

### LINKED ISSUE #24: Implement Support for the Component Class field within the  ComponentIdentifier .

cyrus-dev:
This is new code that parses a new field in the upcoming TCG spec for the platform components fields. The new field indicates the type of hardware (ex Memory - DDR3). This information wasn't provided before so it wasn't always clear what the component was. The new information is provided in a json file. A unit test was created to test the different variations. This commit does not include hooks in the base code to use this class yet. This commit is mainly to include the added library and correct bug and checkstyle issues associated with the new code.

Closes #24


HIRS_Utils/src/main/java/hirs/data/persist/certificate/attributes/ComponentClass.java
```diff
@@ -0,0 +1,262 @@
...
+     */
+    private void getCategory(final JsonObject categories) {
+        int componentID;
+
+        if (categories != null) {
+            for (String name : categories.names()) {
+                componentID = Integer.decode(categories.get(name).asObject().get("ID").asString());
+                // check for the correct flag
+                if ((componentIdentifier & CATEGORY_MASK) == componentID) {
+                    JsonObject subObjects = categories.get(name).asObject().get("Types").asObject();
```
chubtub: Can we name this JsonObject types instead of subObjects? Or something more indicative of what we're storing.

cyrus-dev: Updated the names and also of another variable that was just 'object '.

PR #116, opened 2019-03-27
[#61] Remove Magic String Usage from PersistenceConfiguration

### LINKED ISSUE #61: HIRS_Utils PersistenceConfiguration Should Avoid Magic Strings
`PersistenceConfiguration` in HIRS_Utils subproject should avoid using magic Strings for Spring Bean Names. If anyone references one and it changes, then their code will break unnecessarily. A small refactor should be done to ensure they could reference a dedicated, statically defined String, that if an update occurs, they will also receive the update.

apldev3:
Super simple MR to remove the use of a magic string that originated in the HIRS_Utils PersistenceConfiguration class.

Closes #61


apldev3:
Got a +1, merging now.

PR #118, opened 2019-03-28
[#18] Add in Null Checks for Certificate and Subclasses

### LINKED ISSUE #18: Add in Null Checks for Certificate
In `HIRS_Utils` we have the `Certificate` class in charge of parsing Certificates of all varieties. There is a small problem in that they come in as a byte array that has no guarantees of being of having all the required fields in a parsable format.

The entire `Certificate` class hierarchy should be scanned for locations where this assumption is made and null checks introduced appropriately. Required fields may need to throw an error.

Issue comes out of merge request #16

apldev3:
Small MR to add in a couple null checks to `Certificate` and `PlatformCredential`. I looked over `Certificate` and all of its subclasses pretty thoroughly and I only found a few places that needed an actual null check. Most of the other fields have to be there in order to form a valid Certificate, so it's more likely that either BouncyCastle or the Java Security API would throw a fit before we came across any issues.

With regards to testing, I know we don't have any certificates in violation of these new null checks, and, personally, I'm not sure where to get them. I would think PACCOR could create a new platform credential but I don't know that it has the ability to create one without making any extensions, and I'm not sure that I would request that as a feature. I'm also not certain that it's even possible to create an `AttributeCertificate` without the `AttributeCertificateInfo` field being present, but the BouncyCastle classes didn't seem to enforce it so I threw in the `null` check to be safe.

If y'all have any ideas let me know.

Closes #18

### DIFF
--- a/HIRS_Utils/src/main/java/hirs/data/persist/certificate/Certificate.java
+++ b/HIRS_Utils/src/main/java/hirs/data/persist/certificate/Certificate.java
@@ -11,6 +11,7 @@ import org.bouncycastle.asn1.ASN1Encodable;
 import org.bouncycastle.asn1.ASN1Integer;
 import org.bouncycastle.asn1.ASN1Primitive;
 import org.bouncycastle.asn1.ASN1Sequence;
+import org.bouncycastle.asn1.DERBitString;
 import org.bouncycastle.asn1.x500.X500Name;
 import org.bouncycastle.asn1.x500.style.RFC4519Style;
 import org.bouncycastle.asn1.x500.AttributeTypeAndValue;
@@ -18,6 +19,7 @@ import org.bouncycastle.asn1.x500.RDN;
 import org.bouncycastle.asn1.x509.AttributeCertificate;
 import org.bouncycastle.asn1.x509.AttributeCertificateInfo;
 import org.bouncycastle.asn1.x509.AttCertIssuer;
+import org.bouncycastle.asn1.x509.Extensions;
 import org.bouncycastle.asn1.x509.V2Form;
 import org.bouncycastle.asn1.x509.GeneralNames;
 import org.bouncycastle.cert.X509AttributeCertificateHolder;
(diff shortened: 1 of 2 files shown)


HIRS_Utils/src/main/java/hirs/data/persist/certificate/Certificate.java
```diff
@@ -11,13 +11,15 @@
 import org.bouncycastle.asn1.ASN1Integer;
 import org.bouncycastle.asn1.ASN1Primitive;
 import org.bouncycastle.asn1.ASN1Sequence;
+import org.bouncycastle.asn1.DERBitString;
```
apldev4: Unused import.
```suggestion
```

apldev3: Fixed!

HIRS_Utils/src/main/java/hirs/data/persist/certificate/Certificate.java
```diff
@@ -394,11 +400,16 @@ public Certificate(final byte[] certificateBytes) throws IOException {
...
-                                attCertInfo.getExtensions()));
+                Extensions attCertInfoExtensions = attCertInfo.getExtensions();
+                if (attCertInfoExtensions != null) {
+                    authKeyIdentifier = AuthorityKeyIdentifier
+                            .fromExtensions(attCertInfoExtensions);
+                    this.authorityInfoAccess = getAuthorityInfoAccess(
+                            AuthorityInformationAccess.fromExtensions(
+                                    attCertInfoExtensions));
+                } else {
+                    authKeyIdentifier = null;
```
cyrus-dev: The initialization of the temporary value could be set to null and remove the need for an else statement.

apldev3: Totally fair and fixed.

apldev3:
Every comment has been resolved. Merging and closing.

PR #120, opened 2019-03-28
[#119] tpm_aca_provision no longer tries to guess the TPM version.

### LINKED ISSUE #119: tpm_aca_provision detects a TPM 2.0 device as 1.2 and does not allow a provision to occur
@iadgovuser26 reported:

When issuing a tpm_aca_provision I get the following error
sudo tpm_aca_provision
TPM 1.2 detected.
The package 'HIRS_Provisioner_TPM_1_2' must be installed to provision a TPM 1.2.

However when I look at dmesg I see a 2.0 device:
dmesg | grep TPM
[ 0.000000] ACPI: TPM2 000000007aade618 00034 (v03 DELL CBX3 00000001 AMI 00000000)

When i use hirs-provisioner -p the process works.

apldev4:
The tpm_aca_provision script used to try to guess the version
of the TPM on the machine and call the appropriate command
to provision the TPM. However:
1. The guessing was error prone,
2. The command to provision the TPM is the same regardless of the
TPM version, and
3. By the time the script is run, either the TPM Provisioner 1.2
package or the TPM Provisioner 2.0 package has already been
installed. By that point, it's too late to be wondering which
version of TPM is on the machine. The decision of which package
to install is guided by the documentation.

Closes #119.


apldev3:
@apldev4 are we planning on creating a tpm2_aca_provision script as well?

apldev4:
> are we planning on creating a tpm2_aca_provision script as well?

It's not hard to add that if it is deemed necessary. It can be done in a future issue.

apldev3:
Sounds good to me.

HIRS_Provisioner/scripts/install/tpm_aca_provision
```diff
@@ -1,81 +1,3 @@
...
-		RC=$?
-	else
-		echo "TPM 1.2 provisioner is not installed; not running normal HIRS provisioning."
-		exit 1
-	fi
-fi
-
-exit $RC
\ No newline at end of file
+hirs-provisioner provision
```
apldev3: Real quick is it `hirs-provisioner provision` or `hirs-provisioner --provision`?

apldev4: For the TPM 1.2 Provisioner, it can be `hirs-provisioner -p`, `hirs-provisioner --provision`, or `hirs-provisioner provision`.
For the TPM 2.0 Provisioner, it must be `hirs-provisioner provision`.

I went with the one that works for both.

apldev3: Alrighty, well this all looks fine to me. Feel free to merge if you want. Or we can wait for @iadgovuser26 to give the thumbs up.

PR #125, opened 2019-04-10
[#124] Added length to DigestAlgorithm enum values.

### LINKED ISSUE #124: DigestAlgorithm enum values could know their lengths
`DigestAlgorithm` is an enum that only knows the names of various hashing algorithms. The lengths of those hash values would be good to store in those enum values.

apldev4:
Added length to DigestAlgorithm enum values.

Closes #124.


apldev3:
Excellent refactor. Always like seeing use of Java's support for extended Enum constructors.

PR #126, opened 2019-04-11
[#108] Validity Check for Base and Delta Certs

### LINKED ISSUE #108: Create a Validity Check method for Platform Certs
The V1.1 version of the Platform Certificate Profile added a new concept of "Delta" Platform certificate which can be created after the initial (base) Platform certificate is created.  A new Validity check capability will be needed that collects all the Platform Certificates associated with a device (the "Platform Certificate Chain") and process them to determine if the set is valid or invalid. The actual individual validity checks will be added as separate issues. This issue only address the creation of the method to support individual validity checks. 

The validity check is performed only if the check platform attributes policy is checked.
If the  Platform Certificate set is considered invalid, the processing of the certificate should fail and an attestation certificate should NOT be issued.
If the Platform Certificate set is considered valid then the processing of the certificate can continue.

chubtub:
This is generally how I plan to distinguish and check base and delta certs accordingly.

I decided to assign SupplyChainValidationServiceImpl.baseCredential in the policy.isPcValidationEnabled() loop to avoid looping through the set of PlatformCredential objects again just to find it.  baseCredential is then used to determine in the subsequent policy.isPcAttributeValidationEnabled() loop whether to validate a base or a delta credential, each with their respective methods in SupplyChainCredentialValidator.

Closes #108

### DIFF
--- a/HIRS_AttestationCA/src/main/java/hirs/attestationca/service/SupplyChainValidationServiceImpl.java
+++ b/HIRS_AttestationCA/src/main/java/hirs/attestationca/service/SupplyChainValidationServiceImpl.java
@@ -100,6 +100,7 @@ public class SupplyChainValidationServiceImpl implements SupplyChainValidationSe
                 supplyChainAppraiser);
         boolean acceptExpiredCerts = policy.isExpiredCertificateValidationEnabled();
         HashMap<PlatformCredential, SupplyChainValidation> credentialMap = new HashMap<>();
+        PlatformCredential baseCredential = null;
 
         List<SupplyChainValidation> validations = new ArrayList<>();
 
@@ -137,6 +138,14 @@ public class SupplyChainValidationServiceImpl implements SupplyChainValidationSe
                         pc.setDevice(device);
                         this.certificateManager.update(pc);
                         credentialMap.put(pc, platformScv);
+                        /*
+                         * This method will be added to PlatformCredential to return whether a given
+                         * object is a base or a delta credential.
+                         */
+/*                        if (pc.isBase()) {
+                            baseCredential = pc;
+                        }
+*/
                     }
                 }
             }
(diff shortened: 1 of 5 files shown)


HIRS_AttestationCA/src/main/java/hirs/attestationca/service/SupplyChainValidationServiceImpl.java
```diff
@@ -288,6 +303,41 @@ private SupplyChainValidation validatePlatformCredentialAttributes(final Platfor
         }
     }
 
+    private SupplyChainValidation validateDeltaPlatformCredentialAttributes(
+            final PlatformCredential delta,
+            final DeviceInfoReport deviceInfoReport,
+            final PlatformCredential base) {
+    /*
+     * Do we need a new ValidationType for deltas?
```
busaboy1340: Think a new validation type would be helpful.

HIRS_AttestationCA/src/main/java/hirs/attestationca/service/SupplyChainValidationServiceImpl.java
```diff
@@ -288,6 +303,41 @@ private SupplyChainValidation validatePlatformCredentialAttributes(final Platfor
...
+     */
+        final SupplyChainValidation.ValidationType validationType =
+                SupplyChainValidation.ValidationType.PLATFORM_CREDENTIAL_ATTRIBUTES;
+
+        if (delta == null) {
```
busaboy1340: Should you also check if the deviceInfoReport is null? Not sure how much "defensive programming" is expected/recommended

HIRS_AttestationCA/src/main/java/hirs/attestationca/service/SupplyChainValidationServiceImpl.java
chubtub: Added a new type for delta attributes.  I think the existing ValidationType for (base) platform credentials can extend to deltas as well so I did not create a new enum element for delta platform credentials @cyrus-dev

HIRS_AttestationCA/src/main/java/hirs/attestationca/service/SupplyChainValidationServiceImpl.java
chubtub: The AbstractACA class does a null check on the deviceInfoReport before it is rolled into a Device object and passed to SupplyChainValidationService.

PR #127, opened 2019-04-24
[#105] Updated tpm_version to get data from TPM hardware.

### LINKED ISSUE #105: Rewrite tpm_version from tpm2 provisioner package
With the introduction of the `HIRS_ProvisionerTPM2/src/tpm_version.cpp` binary from #100, default values were provided for detecting the manufacturer and TPM version. This needs to be written to actually collect the manufacturer and version from the TPM rather than providing default values.

apldev4:
The TPM 2 Provisioner gets packaged with an application called
tpm_version, which reported hard-coded values for the TPM
manufacturer and version. Now it collects those things from
the TPM and reports them.

Closes #105.

### DIFF
--- a/HIRS_ProvisionerTPM2/docs/developer-dependencies-centos.md
+++ b/HIRS_ProvisionerTPM2/docs/developer-dependencies-centos.md
@@ -7,20 +7,21 @@ Please look up their respective names in the CentOS repositories.
 
 If no available repository for the development environment contains the dependencies at an acceptable version level, it is expected that the packages be retrieved and built from their respective source repositories.
 
-| Dependency        | Version used | Minimum required   | Repository            | Project repository                     |
-| ----------------- | ------------ | ------------------ | --------------------- | -------------------------------------- |
-| cppcheck          | 1.80         | 1.72               | CentOS 7 epel-release | http://cppcheck.sourceforge.net/       |
-| doxygen           | 1.8.13       | 1.8.0 (estimated)  | CentOS 7 base         | https://github.com/doxygen/doxygen     |
-| graphviz          | 2.30.1       | 2.28.0 (estimated) | CentOS 7 base         | https://gitlab.com/graphviz/graphviz   |
-| gcc-c++           | 4.8.5        | 4.8.5              | CentOS 7 base         | https://gcc.gnu.org/                   |
-| libcurl-devel     | 7.29.0       | 7.0.0 (estimated)  | CentOS 7 base         | https://github.com/curl/curl           |
-| libssh2-devel     | 1.4.3        | 1.4.3 (estimated)  | CentOS 7 base         | https://github.com/libssh2/libssh2     |
-| log4cplus-devel   | 1.1.3        | 1.1.2              | CentOS 7 epel-release | https://github.com/log4cplus/log4cplus |
-| openssl-devel     | 1.0.2k       | 1.0.2g (estimated) | CentOS 7 base         | https://github.com/openssl/openssl     |
-| protobuf-compiler | 2.5.0        | 2.4.1 (estimated)  | CentOS 7 base         | https://github.com/google/protobuf     |
-| protobuf-devel    | 2.5.0        | 2.4.1 (estimated)  | CentOS 7 base         | https://github.com/google/protobuf     |
-| re2-devel         | 20160401     | 20160201           | CentOS 7 epel-release | https://github.com/google/re2          |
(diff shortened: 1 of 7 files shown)


HIRS_ProvisionerTPM2/docs/developer-dependencies-centos.md
```diff
@@ -7,20 +7,21 @@ Please look up their respective names in the CentOS repositories.
...
+| openssl-devel     | 1.0.2k       | 1.0.2g (estimated) | CentOS 7 base         | https://github.com/openssl/openssl          |
+| protobuf-compiler | 2.5.0        | 2.4.1 (estimated)  | CentOS 7 base         | https://github.com/google/protobuf          |
+| protobuf-devel    | 2.5.0        | 2.4.1 (estimated)  | CentOS 7 base         | https://github.com/google/protobuf          |
+| re2-devel         | 20160401     | 20160201           | CentOS 7 epel-release | https://github.com/google/re2               |
+| tpm2-tss-devel    | 1.2.0        | 1.0.0              | CentOS 7 base         | https://github.com/intel/tpm2-tss           |
+| tpm2-abrmd-devel  | 1.1.0        | 1.1.0              | CentOS 7 base         | https://github.com/tpm2-software/tpm2-abrmd |
+| cmake             | 2.8.12.2     | 2.6.0 (estimated)  | CentOS 7 base         | https://cmake.org/                          |
+| cpack             | 2.8.12.2     | 2.6.0 (estimated)  | CentOS 7 base         | https://cmake.org/                          |
+| git               | 1.8.3.1      | 1.6.0 (estimated)  | CentOS 7 base         | https://github.com/git/git                  |
```
apldev6: What has changed in this file?

apldev4: In order to compile `tpm_version`, we now need to install the package `tpm2-abrmd-devel`. The whitespace on this whole table changed to get the formatting to look nice, so it looks like I changed more than I really did.

HIRS_ProvisionerTPM2/src/tpm_version.cpp
```diff
@@ -1,11 +1,125 @@
...
+class SapiContext {
+    TSS2_TCTI_CONTEXT *tctiContext = nullptr;
+    TSS2_SYS_CONTEXT *sapiContext = nullptr;
+
+    SapiContext(TSS2_TCTI_CONTEXT* tcti_ctx, TSS2_SYS_CONTEXT* sapi_ctx)
+            : tctiContext(tcti_ctx), sapiContext(sapi_ctx) {}
+public:
+
+    ~SapiContext() {free(sapiContext); free(tctiContext);}
+    static shared_ptr<SapiContext> create() {
```
apldev2: Can you add some documentation and just some general commenting to this file?

apldev3:
Failed at the Ubuntu packaging. Got it to package locally on my Ubuntu VM after installing the package I think it needs. Will update and push a fix.

apldev3:
Alright the update didn't work, because the package was a) not correct and b) the correct one does not exist for Ubuntu. The reason it worked locally was because I had downloaded and built tpm2-abrmd and installed it from source, which will have to be the solution here as well.

apldev3:
Alrighty, checked it and we're all good at this point. As expected had to build tpm2-tss and tpm2-abrmd from source on Ubuntu to get the library that the `tpm_version` binary is expecting but the Docker image is packaging it fine once again.

apldev3:
This is all good to go @apldev4

apldev3:
For anyone looking I updated the Dockerfiles and the Debian Developer Dependency files with the changes.

HIRS_ProvisionerTPM2/src/tpm_version.cpp
apldev4: will do

HIRS_ProvisionerTPM2/src/tpm_version.cpp
```diff
@@ -1,11 +1,190 @@
...
+/**
+ * Returns the 4-byte string represented by the bytes of value.
+ *
+ * Assumes the bytes of value are reversed.
+ * @param value the 4 bytes to be reversed and placed in the returned string
+ * @return the string represented by value
+ */
+string reversedStringValue(UINT32 value) {
+    string stringValue(4, 0);  // make a string on length 4
+    std::memcpy(&stringValue[0], &value, sizeof value);  // copy bytes in
```
apldev6: Suggest referencing the size of the target versus the source when specifying size for the memcpy.

apldev4: okay.

PR #143, opened 2019-05-20
[#142] Make sure Tomcat stops completely in Docker containers before restarting it

### LINKED ISSUE #142: Restarting Tomcat in Docker Can Have Race Condition
When restarting Tomcat in Docker, we have to take a more manual approach to stopping and starting the server as opposed to a quick `service tomcat restart` or similar approach. This goes back to the issues with the lack of DBus in Tomcat, but the workaround seems to be fine it most instances unless you layer another webapp on top of this one.

This causes Tomcat to take longer to stop than expected, so when it attempts to start a new instance of Tomcat immediately after invoking the stop command, then it causes a Port Binding collision.

Current proposed solution is to introduce a sleep between the stop/restart of the Tomcat server (for the Docker config) to make sure it has time to clean everything up. This shouldn't be an issue outside Docker since normal systems should have access to the appropriate service management tools.

apldev3:
This is a small MR that does as advertised. We want to avoid port collisions when the ACA is run parallel with other webapps in a Tomcat deployment in Docker containers.

Closes #142

### DIFF
--- a/package/scripts/aca/certificate_generate.sh
+++ b/package/scripts/aca/certificate_generate.sh
@@ -75,6 +75,10 @@ if [ $DOCKER_CONTAINER = true ]; then
     if [[ $(ss -t -l -n | grep -q LISTEN.*:::8009) -eq 0 ]]; then
         echo "Tomcat is running, so we restart it."
         /usr/libexec/tomcat/server stop
+        # Wait for Tomcat to stop completely and prevent port bind collisions
+        while [ -z $(tail -n 1 /var/log/tomcat/catalina.$(date +"%Y-%m-%d").log | grep "Destroying ProtocolHandler \[\"http-bio-8443\"\]") ]; do
+            :
+        done
         (/usr/libexec/tomcat/server start) &
         # Wait for Tomcat to boot completely
         until [ "`curl --silent --connect-timeout 1 -I http://localhost:8080 | grep 'Coyote'`" != "" ]; do
--- a/package/scripts/aca/certificate_generate.sh
+++ b/package/scripts/aca/certificate_generate.sh
@@ -76,7 +76,7 @@ if [ $DOCKER_CONTAINER = true ]; then
         echo "Tomcat is running, so we restart it."
         /usr/libexec/tomcat/server stop
         # Wait for Tomcat to stop completely and prevent port bind collisions
-        while [ -z $(tail -n 1 /var/log/tomcat/catalina.$(date +"%Y-%m-%d").log | grep "Destroying ProtocolHandler \[\"http-bio-8443\"\]") ]; do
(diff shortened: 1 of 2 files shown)


package/scripts/aca/certificate_generate.sh
```diff
@@ -75,6 +75,10 @@ if [ $DOCKER_CONTAINER = true ]; then
     if [[ $(ss -t -l -n | grep -q LISTEN.*:::8009) -eq 0 ]]; then
         echo "Tomcat is running, so we restart it."
         /usr/libexec/tomcat/server stop
+        # Wait for Tomcat to stop completely and prevent port bind collisions
+        while [ -z $(tail -n 1 /var/log/tomcat/catalina.*.log | grep "Destroying ProtocolHandler \[\"http-bio-8443\"\]") ]; do
```
apldev4: There are two things to be concerned about here:
1. Because it checks all log files in the directory, the `while` condition is going to tell us if _any_ log file exists from a time when tomcat was stopped. Isn't that too broad? We want to know if it is stopped now -- not ever.
1. The `$()` needs double quotes around it or there will be an error message about "too many arguments". (`-z` only takes a single argument.)

apldev3: 1) This is only ever going to run when the ACA is deployed in a Docker container. Presumably each new instance of the Docker container will install the ACA on startup, so it's unlikely there should ever be anymore than the current day's worth of catalina.*.log. That being said, if you know of a good way to get today's date in the catalina format we can update that.

2) It doesn't appear to need double quotes, and it seems to be running just fine. If it didn't then my Docker containers would get trapped in this while loop or at the very least throw the error you suggest.

apldev3: I just pushed an update that should resolve your first comment.

package/scripts/aca/certificate_generate.sh
```diff
@@ -75,6 +75,10 @@ if [ $DOCKER_CONTAINER = true ]; then
     if [[ $(ss -t -l -n | grep -q LISTEN.*:::8009) -eq 0 ]]; then
         echo "Tomcat is running, so we restart it."
         /usr/libexec/tomcat/server stop
+        # Wait for Tomcat to stop completely and prevent port bind collisions
+        while [ -z $(tail -n 1 /var/log/tomcat/catalina.$(date +"%Y-%m-%d").log | grep "Destroying ProtocolHandler \[\"http-bio-8443\"\]") ]; do
```
apldev4: The argument passed to `-z` needs to be a single string or you will get an error. You need to wrap it in double quotes.
```suggestion
        while [ -z "$(tail -n 1 /var/log/tomcat/catalina.$(date +"%Y-%m-%d").log | grep "Destroying ProtocolHandler \[\"http-bio-8443\"\]")" ]; do
```

PR #148, opened 2019-05-22
Updated version to v1.0.4

iadgovuser26:
Updated version file.


iadgovuser26:
Deleted the existing V1.0.4 tag and renamed the V104 to V1.0.4 tag.

apldev3:
Might still want to delete the v104 tag.

image

Issue #176, opened 2019-08-01
Add Delta system test for initial Base certificate system test.


busaboy1340:
Add delta system test that resolves the bad base certificate test from issue #132

busaboy1340:
Discovered possible new ACA issue. When provisioning with a delta that has a component without a serial number, the ACA rejects it with error: "There are errors with Delta Component. Status component: was ADDED, the serial already exists".
According to the spec, the componentSerial is an OPTIONAL field, but the aca is treating it as REQUIRED.

busaboy1340:
Closed by PR #187

PR #177, opened 2019-08-01
[#176] Initial delta system test to resolve bad component in a base certificate

busaboy1340:
Initial delta system test to resolve bad component in a base certificate.


.ci/system-tests/system_test_Driver.py
```diff
@@ -0,0 +1,477 @@
...
+provisioner_out = None
+
+HIRS_ACA_PROVISIONER_IP="172.19.0.3"
+HIRS_ACA_PROVISIONER_TPM2_IP="172.19.0.4"
+TPM_ENABLED=True
+IMA_ENABLED=False
+
+# Change accordingly
+#HIRS_ACA_PORTAL_IP="172.17.0.2"
+HIRS_ACA_PORTAL_IP="rd8ul-31673ab.dod.mil"
```
cyrus-dev: I thought we weren't going to load system specific information.

busaboy1340: Your right. This is just a driver I use to make python debugging easier. We can just get rid of the file itself.

PR #183, opened 2019-08-22
[#180] Upgrade hirs_db schema according to upgrade version

### LINKED ISSUE #180: Add support for yum update from previous releases to v1.1.0.
Yum update for the v1.1.0 candidate release ACA rpm package does not currently account for database updates made to support for Delta and Base certificates. DB changes were made for release 1.0.4 and release 1.1.0. The RPM installation should check for previous versions and run the appropriate db scripts to accommodate db changes made on these versions.

chubtub:
At the time of writing there are schema changes when upgrading from 1.0.3 to 1.0.4 but not when upgrading to 1.1.0.  These scripts are called when 'yum upgrade HIRS_AttestationCA' is invoked.

Closes #180


chubtub:
This will work for upgrading (1.0.3 to 1.0.4) and (1.0.4 to 1.1.0) but not (1.0.3 to 1.1.0) directly.  I'm modifying the sql script to check for the existence of the columns before adding them.

package/scripts/common/upgrade_schema_1.0.4.sql
```diff
@@ -0,0 +1,10 @@
+ALTER TABLE Certificate
+ADD isDeltaChain bit(1) DEFAULT NULL,
+ADD platformBase bit(1) DEFAULT NULL,
```
busaboy1340: Should this be "isPlatformBase" ?

chubtub:
rpm -U or yum upgrade now updates both the db schema and the banner version number.

Issue #181, opened 2019-08-13
Base Certificate Fails after Delta status fix


cyrus-dev:
During #166 the code was modified to redo how the validations are stored.  This has the effect of a faulty base certificate showing fail when a delta certificate was uploaded.  The code has to be reworked to ignore the attributes failure on the base if there are delta certificates.

PR #184, opened 2019-08-23
[#181] Attribute Validation updated

cyrus-dev:
The validation page was not showing an error icon for attributes failures.  This was due to the retained validation type for attributes.  This has been removed and the code was additionally updated with logic to handle showing just one icon for both policy checks for the platform credential.

Associated with #181


cyrus-dev:
As an additional note, the unit test was flipped because there is no case in which the pc attributes are validated without the general pc validations are done.

Issue #222, opened 2020-01-28
Add processing for the TCG Event Log


iadgovuser26:
One of the Support RIM formats supported by HIRS will be the TCG Event Log as described in the the Platform Firmware Profile. The ACA will need to be able to parse the TCG log and produce a set of expected PCR Values which HIRS internally stores in the "TPM Baseline" table. The TPM Baseline gets used for processing a TPM quote during the ACA processing of and identity request message if the ACA policy requires it. The new class will be added  to HIRS_UTILS and provide the following capabilities:

1. Parsing the Event Log and extracting expected PCR values from the log. 
2. Storing a TPM Whitelist Baseline generated from a TCG Event Log.
3. Provide TPM Values for display purposes on the ACA portal.

Note that signature verification of the Base RIM and subsequent digest verification on the support RIM should occur prior to the invocation of this class.

iadgovuser26:
merged by PR  #223.

PR #223, opened 2020-02-04
[222] Added TCG Event Log Processing that converts TCG Event Logs to HIRS T…

iadgovuser26:
Adds a hirs.tpm.eventlog package to HIRS_Utils in support of issue #222 . 
The TCGEventLogProcessor class will parse a TCG Event log. Capability added includes:
1.  Produces the "expected PCRs" (PCR Assertions) from the TCG Event log.
2. Provides a single array representation of all 24 expected PCRs for display purposes.
3. Provides a single string representation of a single expected PCR for display purposes. 
4.  Creates a TpmWhiteListBaseline from the TCG Event Log which can be used with the HIRS TPM Appraiser.
5. Supports both the  "SHA1Format" and "Crypto Agile Format" defined by the TCG PC Client Platform Firmware Profile.  

This functionality is in support the processing of the TCG Reference Integrity Manifest (RIM).


HIRS_Utils/src/main/java/hirs/utils/HexUtils.java
```diff
@@ -74,4 +76,41 @@ public static Integer hexToInt(final String s) {
         System.arraycopy(b, start, copy, 0, end - start + 1);
         return copy;
     }
+
+    /**
+     * Takes in a byte array and reverses the order.
+     * @param in  byte array to reverse
+     * @return reversed byte array
+     */
+   public static byte[] leReverseByte(final byte[] in) {
```
cyrus-dev: You can use ArrayUtils.reverse(byte[]) or Collections.reverse(List<T>).

HIRS_Utils/src/main/java/hirs/tpm/eventlog/TpmPcrEvent.java
```diff
@@ -0,0 +1,178 @@
...
+  /** Length (in bytes) of a pcr. */
+  private int digestLength = 0;
+
+  /**
+  * Constructor.
+  * @param is ByteArrayInputStream holding the event
+  * @throws IOException when event can't be parsed
+  */
+  public TpmPcrEvent(final ByteArrayInputStream is) throws IOException {
+
```
cyrus-dev: I was looking at this and wondering why you needed an empty constructor that the children call, but the parameter isn't stored or used.

HIRS_Utils/src/main/java/hirs/tpm/eventlog/TpmPcrEvent1.java
```diff
@@ -0,0 +1,49 @@
...
+      setPcrIndex(unit32data);
+      is.read(unit32data);
+      setEventType(unit32data);
+      byte[] eventdigest = new byte[TpmPcrEvent.SHA1_LENGTH];
+      is.read(eventdigest);
+      setDigest(eventdigest);
+      is.read(unit32data);
+      int eventSize = HexUtils.leReverseInt(unit32data);
+      byte[] eventContent = new byte[eventSize];
+      is.read(eventContent);
```
cyrus-dev: eventContent isn't stored.  Is that ok?  TpmPcrEvent2 stores a like variable in its constructor.

HIRS_Utils/src/main/java/hirs/tpm/eventlog/TpmPcrEvent.java
iadgovuser26: Right now the Event log is only being processed for expected PCR values. In a future PR we will be adding capability to process individual events. This is not technically needed right now but is placed here in anticipation of that update.

HIRS_Utils/src/main/java/hirs/tpm/eventlog/TpmPcrEvent1.java
iadgovuser26: Same reasoning as the previous comment, its intended to be used inn a future PR. Just wanted to get the event content ready for future processing. Storing it wont be necessary as the content is envisioned to be used for temporary processing.

HIRS_Utils/src/main/java/hirs/utils/HexUtils.java
iadgovuser26: OK, will replace in a future PR.

cyrus-dev: I added an issue to cover this. #227

Issue #231, opened 2020-02-25
tcg_rim_tool CLI changes


chubtub:
This is a follow on to #217 to add more features to the command line interface.

PR #232, opened 2020-02-25
[#231] Last minute tcg_rim_tool CLI changes

chubtub:
Added command line argument to handle visibility of signing certificate.  The default is to not show `<X509Certificate>` in the signature block of the generated base RIM.  `<X509Certificate>` is visible with the presence of `--show-cert`.


tools/tcg_rim_tool/src/main/java/hirs/swid/utils/Commander.java
```diff
@@ -106,10 +106,15 @@ public final void parseArguments(final String[] args) {
                 	parse = true;
                 	parseFile = args[++i];
                 	break;
-                case FULL_COMMAND_PREFIX + KEY_STRING:
+/*
```
busaboy1340: Guess you can get rid of this commented out code?

chubtub: Comments deleted.

PR #246, opened 2020-04-03
HexUtils Refactor

cyrus-dev:
The HexUtils class has a lot of code that is unnecessary.  The functionality can be done by built in Java classes, already loaded, that can be used to do the operations HexUtils handles.


cyrus-dev:
At this time, this is being put in the backlog

PR #255, opened 2020-06-05
[#254] Modified how TPM2 Provisioner pulls down sub module cpr

### LINKED ISSUE #254: [BUG] CPR fails for TPM2Provisioner
At some point in the last 2 weeks, there was a change either with the cpr-download project.  Because of this change, the build would fail 

error: Updating the following directories would lose untracked files in it:
	opt/mongoose

This seems unrelated to our specific code changes. Previous builds that passed now fail, going back for months.

cyrus-dev:
This issue was resolved by modifying the CPR.CMakeLists.txt.in to no longer use the GIT_REPOSITORY and GIT_TAG.  Instead the URL and URL_HASH are used, pulling down the zip file.

I'm unsure as to what specifically causes this to work and not the repo pull but I am under the assumption that the zip contains everything that is needed whereas the repo and the git commands create the issue that it can't resolve when trying to pull the needed libraries.

Closes #254

### DIFF
--- a/HIRS_ProvisionerTPM2/lib/CPR.CMakeLists.txt.in
+++ b/HIRS_ProvisionerTPM2/lib/CPR.CMakeLists.txt.in
@@ -4,8 +4,8 @@ project(cpr-download NONE)
 
 include(ExternalProject)
 ExternalProject_Add(cpr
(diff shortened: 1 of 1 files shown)


cyrus-dev:
This problem originated in travis.  At this time the branch now passes successfully.  Merging.

PR #262, opened 2020-06-12
[#257] Changes for 2.0 beta

### LINKED ISSUE #257: RIM tool changes for 6/12 release
Two changes will be implemented by the end of this week:
1. Disallow the embedding of self-signed certificates in base RIMs
2. Require one support RIM and calculate its hash to populate the payload of the base RIM. This should then be verified during --verify.

chubtub:
- Added a check against embedding a self-signed cert via -p option.
- Removed requirement of a public cert for -v since currently only verification by embedded cert is supported.
- Added requirement of a support rim whose name, size, and hash are added to the payload.

Closes #257


iadgovuser26:
:+1:

PR #276, opened 2020-06-25
[#275] Modify Payload filepath

### LINKED ISSUE #275: Payload file path
The current version of the RIM tool places the entire relative filepath from `-l <path>` in the name attribute of the <File> under <Payload>.  However the default file location of rimel files should be /boot/manifest/rim, which is not where test files are always placed.  This is causing issues when importing test patterns to the ACA.

chubtub:
1. <File> under <Payload> for a rimel has name read from the attributes json file, while size and hash are calculated from the file given through `-l` command-line parameter.
2. `-v` will verify the <File> hash using what is given in the base RIM; `-v -l <filepath>` will verify the <File> hash using <filepath> instead.

Closes #275


iadgovuser26:
:+1:  Works with PR 277!

PR #277, opened 2020-06-25
[#274] Swid Resource Not Found

### LINKED ISSUE #274: Update appearance of RIM when resource not present
Currently when the ACA only have the RIM and no associated resource (event log) file the details page shows a "blank" list of expected PCRs.

This change will instead list the name of the file that is missing.

cyrus-dev:
Updated RIM Details page to display File Not Found when the associated event log has not been uploaded with the swid tag.

Closes #274


iadgovuser26:
"File not found" shows up in the Support RIM section under the Directory /Files/filename. It may not be so obvious that this relates to the upload of the Support RIM. Can there be a tooltip on the file not found that states something like "Support RIM file named **** was not imported via the Reference Integrity Manifest page" where **** is the filename.

iadgovuser26:
Can we add the term "Payload" to the section that is currently labelled "Support RIM" as this would be consistent with the use of SWID tag element names in that column.

Issue #298, opened 2020-08-28
Base RIM signature verification


iadgovuser26:
The Base RIM needs to be able to validate the signature on the Base RIM. This implies that the Trust Chains certificate used for the Base RIM verification have been previously loaded into the Trust Store via the ACA.s Trust Chain Management Page.

1. Signature verification check of the Base Rim needs to be added to the firmware validation process. This needs to include a check of the entire certificate path.
2. The Base RIM details page needs a signature section to include a signer row with a link to the signing certificate.
   a. A green check box in the signature section  to indicate the signature is valid. 
3. The hash of the support RIM needs to be compared against the payload file hash within the Base RIM.

PR #300, opened 2020-09-18
[#298] Add base RIM validation to frontend

chubtub:
Changes can be summarized in the following:

- Added tcg_rim_tool as a dependency for HIRS_AttestationCAPortal
- tcg_rim_tool performs all validation and passes results to ACAPortal
- Added three placeholders to rim-details.jsp: one for payload file validation, one for signature validation, and one for the signing certificate link


HIRS_Utils/src/main/java/hirs/data/persist/ReferenceManifest.java
```diff
@@ -223,7 +228,11 @@ public ReferenceManifest(final byte[] rimBytes) throws IOException {
 
         this.rimBytes = rimBytes.clone();
 
-        SoftwareIdentity si = validateSwidTag(new ByteArrayInputStream(rimBytes));
+        SwidTagValidator validator = new SwidTagValidator();
+        validator.validateSwidtagInputStream(new ByteArrayInputStream(rimBytes));
+        SoftwareIdentity si = validator.getSoftwareIdentity();
+        payloadFileValid = validator.isPayloadFileValid();
```
cyrus-dev, line 234: If this is to compare the file hash in the xml swidtag, please look at the code at the bottom of the constructor.  This is already stored in rimHash.  This is mainly a note for later but going down into SwidTagValidator to compare this is unnecessary and non-functional with the changes I'm making.  The main problem is that this constructor is only for base RIMs and the validation won't work if the support rim isn't uploaded first, which isn't a guarantee.

chubtub:
Superseded by #311

PR #304, opened 2020-10-06
Updated XJC implementation/integration

cyrus-dev:
This is a change to exclude the jaxb generated files from the git tracked file list just like the protobuf class is.  The script runs and generates, the files are ignored by pmd and compile depends on building the xjc directory successfully.


chubtub (approved):
Changes look good

iadgovuser26 (approved):

Issue #238, opened 2020-03-10
Provide TCG Event Log details associated with a PCR mismatches


iadgovuser26:
Issue #222 added processing of the TCG Event Log and provided the ability to create a set of expected PCRs from the Support RIM file.
Issue #236 adds the ability to process a TPM Quote and determine if a PCR value doesnt match an expected value.

What's needed is some human readable details that help explain what a  PCR mismatch implies (stating PCR 4 failed doesn't mean much to most humans).

To support this feature an ability to compare the Event Log from the Provisioner against the Event Log supplied as a Support RIM file for a given PCR is required . The Event Log comparison will provide details about why the PCRs did not match.

The return value of this method will be a string array containing one human readable entry for each event associated with the specific PCR  that had digests that mismatched. The human readable string is based upon the details provided by a particular Event (the Event Content field), in which the information may be very limiting (typically based upon UEFI structures), but still more information that just a PCR number.

cyrus-dev:
At the moment this is being broken into at least 2 to 3 tasks/PRs.  This is to reduce the number of files changed in one PR and to establish goals to hit before moving on to a different objective necessary to complete the issue.
Task 1 - Collect the locally stored files on the provisioner, the rim and swidtag files.  While this issue is specifically about the rim file, the swidtag is just as easy to include.  The processing of the swidtag will be saved for later.

Task 2 - The next step will be to compare the client rim with the uploaded rim during a failure and find the mis-match.

Task 3 - If necessary, this will be to display this information which will depend on the type of information from the compare is requested to display.

PR #307, opened 2020-10-19
[#238] Client Upload rewrite and validation

cyrus-dev:
This commit includes a completed rewrite of the ReferenceManifestSelector framework.  Like the previous rewrite, it was easier and made more sense to create addition classes and that are specific to a type of RIM (base, support, measurement) for referencing in the DB.  Once this was rewritten the code was modified to validate the measurement against the support rim.

This PR is part 2 of #238

### DIFF
--- a/HIRS_Utils/src/main/java/hirs/data/persist/BiosMeasurements.java
+++ b/HIRS_Utils/src/main/java/hirs/data/persist/BiosMeasurements.java
@@ -19,31 +19,31 @@ import java.util.Collection;
 
 /**
  * Sub class that will just focus on PCR Values and Events.
+ * Similar to {@link hirs.data.persist.SupportReferenceManifest}
+ * however this is the live log from the client.
  */
 @Entity
-public class SupportReferenceManifest extends ReferenceManifest {
-    private static final Logger LOGGER = LogManager.getLogger(SupportReferenceManifest.class);
+public class BiosMeasurements extends ReferenceManifest {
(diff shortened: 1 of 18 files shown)


HIRS_Utils/src/main/java/hirs/data/persist/BiosMeasurements.java
```diff
@@ -0,0 +1,185 @@
...
+import java.util.Arrays;
+import java.util.Collection;
+
+/**
+ * Sub class that will just focus on PCR Values and Events.
+ * Similar to {@link hirs.data.persist.SupportReferenceManifest}
+ * however this is the live log from the client.
+ */
+@Entity
+public class BiosMeasurements extends ReferenceManifest {
```
iadgovuser26, line 26: Linux calls the file this relates as binary_bios_measurements so the name BiosMeasurements is understandable, however the term BIOS is outdated and the TCG PC Client PFP refers to data being processed as an Event Log (although there is already an EventLog class). Microsoft refers to it as an Measured Boot Log to make things even more tangled. Perhaps something a little more like TcgMeasurements  or EventLogMeasurements might fit here.

cyrus-dev: Agreed, and updated.

iadgovuser26 (approved):

chubtub (approved):
Changes look good

PR #309, opened 2020-10-29
[#169] AIC policy rule

### LINKED ISSUE #169: Allow ACA to perform Validations without issuing a AIC
Allow ACA to perform Validations without issuing a AIC (Provide automated AIC update as well):
    a. Allow admins to set the Validity period (currently fixed at 10 years). Add to Policy Page.
    b. Allow admins to set AIC update window in which a new AIC will be issued when a provisioning request is made and the existing AIC (in the ACA DB) will expire within that window. Add to Policy Page?
    c. The ACA needs to check if an AIC request is not within the window that the AIC fields required for the current policy settings are missing (e.g. Require A EC validation is set , but no TPM Manufacturer or Model number is in the current AIC.). This could be controlled by a new policy setting"
    d. May need to update client to support new command for system check or add logic to skip AIK generation.

cyrus-dev:
This set of code adds a additional policy setting that allows for the user to define if an Attestation Issued Certificate is created each time a provision is run and in addition the ability to define a validity date to expire on.

Closes #169

Removes #258


chubtub (approved):
Changes look good

PR #312, opened 2020-11-06
[#238] Client log mismatch display

### LINKED ISSUE #238: Provide TCG Event Log details associated with a PCR mismatches

cyrus-dev:
In reference to issue #238 
Task 3 - If necessary, this will be to display this information which will depend on the type of information from the compare is requested to display.

- Updated validation page tool tip for mismatch events, just list number of events
- The validation error icon links to a page that displays just the mismatch

Closes #238


HIRS_AttestationCAPortal/src/main/webapp/WEB-INF/jsp/rim-details.jsp
```diff
@@ -57,7 +57,14 @@
...
                                     <c:forEach items="${initialData.events}" var="event">
-                                        <tr>
+                                        <c:choose>
+                                            <c:when test="${event.isError()}">
+                                                <tr style="background: tomato">
+                                                </c:when>
+                                                <c:otherwise>
+                                                <tr>
+                                                </c:otherwise>
+                                            </c:choose>
```
chubtub, line 67: Nit: whitespace

chubtub (approved):
Changes look good

HIRS_AttestationCAPortal/src/main/webapp/WEB-INF/jsp/rim-details.jsp
cyrus-dev: I guess you mean the format of the lines?

PR #317, opened 2020-11-18
[#316] Update component failure highlight

### LINKED ISSUE #316: Platform Component Highlight
Upon testing of updated code for the ACA and it was noted that the ACA no longer highlights the component(s) that are a part of the validation failure.

The reason for the missing feature has to do with the components being tested did not have serial numbers.  That attribute is what the ACA used to identify failed components. 

This feature needs to be reworked.  The serial number attribute is not a required field.

cyrus-dev:
Updated the code to use the manufacturer together with the model attribute of the component into an identifier when a failure is identified instead of the serial number.

Closes #316


chubtub (approved):
Changes look good

PR #328, opened 2020-12-23
Validation report file

### LINKED ISSUE #281: Validation Report File
Add a download link next to the title on the validation page:

The download should save a CSV file with the following information. A dialogue box may capture a few details about the report not found in the ACA DB:

* Company devices where shipped from
* Contract#
* Report for Date range (default to current date) 

The CSV file should contain those data items as well as: 
Report generation date: (current data/time)

Components Checked as part of these tests:
List Class, CPU, Make, and Model by components in the Platform Certs. Should typically be:
Platform , baseboard (motherboard), CPU, Disks, Memory, NIC Lan, NIC Wireless, and Firmware    

Results (1 row per device)

| Verified Manufacturer | Model| SN |Verification Date| (8)Component Status (P/F)|Device Status (P/F) |  
|-----------------------|------|----| ----------------|--------------------------|--------------------|

Where:
* Verified Manufacturer is the Platform Vendor
* Model is the Platform Model
* SN is the Chassis SN
* Verification Data is the not before time on the Attestation Certificate
* Component Status column is 8 component classes names listed above
* Component Status data is taken from the pass/fail status of the report summary
* Device Status is the overall pass/fail of the report summary

chubtub:
These changes add a download link to the validation report page.  This link displays a dialog to collect some user input, pulls data corresponding to the validation report, and combines all the above into a file written locally to the client.

Closes #281

### DIFF
--- a/HIRS_AttestationCAPortal/src/main/java/hirs/attestationca/portal/page/controllers/ValidationReportsPageController.java
+++ b/HIRS_AttestationCAPortal/src/main/java/hirs/attestationca/portal/page/controllers/ValidationReportsPageController.java
@@ -6,6 +6,9 @@ import hirs.attestationca.portal.datatables.OrderedListQueryDataTableAdapter;
 import hirs.attestationca.portal.page.PageController;
 import hirs.attestationca.portal.page.params.NoPageParams;
 import hirs.data.persist.certificate.Certificate;
+import hirs.data.persist.certificate.PlatformCredential;
+import hirs.data.persist.certificate.attributes.ComponentIdentifier;
+import hirs.persist.CertificateManager;
 import org.apache.logging.log4j.Logger;
 import static org.apache.logging.log4j.LogManager.getLogger;
 import org.hibernate.Criteria;
@@ -16,6 +19,7 @@ import org.springframework.stereotype.Controller;
 import org.springframework.ui.Model;
 import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.bind.annotation.RequestMethod;
+import org.springframework.web.bind.annotation.RequestParam;
 import org.springframework.web.bind.annotation.ResponseBody;
 import org.springframework.web.servlet.ModelAndView;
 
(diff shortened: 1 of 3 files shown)


chubtub:
Two notes:
1. The modal dialog that opens when the download link is clicked does not close when the file is saved.  I believe this is because the download link is created in the javascript function rather than processed in the jsp server side.  Either the download link needs to be moved out of the datatable or the modal dialog needs to be handled differently.
2. It seems a different format might be more appropriate for the file.  Depending on how much data should be pulled for each component identifier the Component Status column might look better in JSON format, for example.

HIRS_AttestationCAPortal/src/main/java/hirs/attestationca/portal/page/controllers/ValidationReportsPageController.java
```diff
@@ -97,4 +114,84 @@ public void modify(final Criteria criteria) {
...
+                "attachment;filename=\"" + deviceName + "_validation_report.txt\"");
+        BufferedWriter bufferedWriter = new BufferedWriter(
+                new OutputStreamWriter(response.getOutputStream(), "UTF-8"));
+        Enumeration parameters = request.getParameterNames();
+        while (parameters.hasMoreElements()) {
+            String parameter = (String) parameters.nextElement();
+            bufferedWriter.append(parameter + ": " + request.getParameter(parameter) + "\n");
+            LOGGER.info(parameter + ": " + request.getParameter(parameter));
+        }
+//        String columnHeaders = "Company, Contract Number, Date Range, Verified Manufacturer, "
```
cyrus-dev, line 155: commented code, do you mean to leave this here?

chubtub: Yes, these are the column headers in case we keep the csv format

chubtub:
CSV was intended for import into a common speadsheet app. Not sure if there updated/universal support for json into speadsheets. Definitely should not be left as a text file.
Should only be one Download icon, preferably on the top of the page, next to the page title.
Contract # seems to add the contract # in the report. Good enough for now. May need to get that from another source at some point.
if I leave the company name blank it should collect reports for all devices. It seem to just get the first device. The Name of the for a blank company Probably Should be AllDevices.
If I enter the company of the second device the report name is fro the first device. Only the company name seems to be correct. The rest of the fields seem to come from the first device.
3:50 PM


6, Date range should have a start and end date. currently seems to just place that field in the report.

chubtub:
Added feedback into changes: 
- Company field adds company name to filename.
- Contract field adds contract number to file.
- Date range filters device reports by timestamp.
- Modal dialog still does not close on submit.

Issue #308, opened 2020-10-22
ACA fails if issuer subject does not include organization (O=)


mwiseman32:
Testing using the swtpm (https://github.com/stefanberger/swtpm) ACA fails because it is requiring an Organization in the Subject even though the Organization is not printed as part of the subject, only the cn.
Error from HIRS_AttestationCA.log:
2020-10-22 12:36:02 [hirs.attestationca.rest.AttestationCertificateAuthorityExceptionHandler.handleExpectedExceptions] ERROR : The ACA has encountered an expected exception: organization cannot be null or empty.

Tested using:
Fedora 32 as host OS
KVM
installed swtpm-tool from standard repo
Tested using sw TPM 1.2 but no reason to expect also not applicable to sw TPM 2.0.

Work around is to force swtpm-tools to have an O= in subject name.
In /usr/share/swtpm/swtpm-localca locate the lines:

- echo "cn=swtpm-localca-rootca" > "${tmp}
and
- echo "cn=swtpm-localca" > "${tmp}"

Add this line under **each**:
- echo "organization=abc" >> "${tmp}"
With abc being any string

iadgovuser26:
The [Ek Cert Profile](https://trustedcomputinggroup.org/wp-content/uploads/Credential_Profile_EK_V2.1_R12_PUBLIC_REVIEW.pdf) , nor the certificate profile it adhears to, [rfc 5280](https://tools.ietf.org/html/rfc5280), have any required attributes on the DN within the certificate, so it is valid to have a Issuer DN without an C, O, OU, even though most certificates have these fields by convention. The Authority Key Identifier (AKI) is intended to be used as the lookup for the certificate, however the ACA is currently using the organization attribute to lookup the issuers cert. 

The [EK Certificate specification](https://trustedcomputinggroup.org/wp-content/uploads/Credential_Profile_EK_V2.1_R12_PUBLIC_REVIEW.pdf) requires the AKI, so it can be used in place of the Issuer DN for certificate lookups.  For the ACA this a fairly low level change that effects all certificate look ups. The Platform Certificate and Attestation Certificate will need to include the use of the AKI. 

The [TCG Platform cert specification](https://trustedcomputinggroup.org/wp-content/uploads/IWG_Platform_Certificate_Profile_v1p1_r19_pub_fixed.pdf) states in section 3.2.1.1 Authority Key Identifier:

 "Assign the value of “subject key identifier” from the issuer’s public key certificate, if available, else omit".

[...]

PR #331, opened 2021-01-14
[#308] Updated unit tests

cyrus-dev:
These are unit tests that were not run on the merge request #327 for issue 308.  The unit tests uses certificate (EC and 2 CAs) that don't match for SKI to AKI or issuer string fields.  The O= field isn't required and the code in #327 was changed to use SKI first then a sorted RDN list.  See #327.  A note was left in the unit tests that the test certs need to be updated.

Another set of unit tests were updated because the passed in null parameter for a set<objects> was throwing NPEs.


chubtub (approved):
Changes look good

PR #339, opened 2021-03-11
[#336] Script to download ACAPortal validation reports

### LINKED ISSUE #336: Validation Report File request.
Issue #218 introduced the Validation Report File.  
What's needed is a curl command to trigger the request from a command line.
Command should include option for the date range, otherwise return all validation reports.

chubtub:
This pull request creates a command line script to invoke the download function on the validation report page on the ACA Portal.  Execution is summarized as:
1. get user parameters for filter start and end times in the format yyyy-mm-dd
2. pull data from the validation-reports/list end point
3. parse data for create time of each report and corresponding device name
4. pass user parameters, create times, and device names to validation-reports/download endpoint
5. CSV output is printed to standard output

Closes #336

### DIFF
--- a/scripts/download_validation_reports.sh
+++ b/scripts/download_validation_reports.sh
@@ -3,8 +3,15 @@
 #User input parameters:
 #$1 filter start date 'yyyy-mm-dd'
 #$2 filter end date 'yyyy-mm-dd'
+#$3 ACA address, default is localhost if not given
 
-endpoint="https://localhost:8443/HIRS_AttestationCAPortal/portal/validation-reports"
+if [ -z "$3" ]
+ then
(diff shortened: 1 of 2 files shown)


scripts/download_validation_reports.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
+#User input parameters:
+#$1 filter start date 'yyyy-mm-dd'
+#$2 filter end date 'yyyy-mm-dd'
+
```
iadgovuser26, line 6: Can an ACA address be added as an optional additional parameter with a default to localhost?

chubtub: The script now looks for a third argument to use as the ACA address, defaulting to localhost if not found.

PR #340, opened 2021-03-17
RIMEL Upload bug fix

cyrus-dev:
This is a change to the details page handler for the RIM files.  The rimel files that are uploaded by themselves are not updated with manufacturer or model information, and the controllers are referencing that information when looking up database information.  This causes a null exception and a funky output on the screen.  This commit fixes this issue.


iadgovuser26 (approved):
:+1:

PR #342, opened 2021-03-19
[#341] RIM Event Digest Store

### LINKED ISSUE #341: Support RIM Event Digest storage
This issue is a preparation issue that takes the events provided in a Support RIM that has a baseline for the measured events expected from the Event Log.

This issue is to make a more direct link between the device manufacturer, model and the digests in the event log for the verification, updates and deletions via supplemental/patch RIMs during the life cycle of the device.

cyrus-dev:
This implementation code that stores the digest from an event log along with other relevant information based on the manufacturer and model.

This Closes #341


chubtub (approved):
Changes look good

iadgovuser26 (approved):

PR #347, opened 2021-03-30
Serial Number Display Fix

cyrus-dev:
Updated the code for authority serial number and serial number for the general info for any certificate to no longer use the getLong method of the Big Integer object.  This truncated the value.  Instead, the new code uses the Hex.toHexString method from bouncy castle.


iadgovuser29 (approved):
It works!

Issue #343, opened 2021-03-22
Validate Firmware Event Digest


cyrus-dev:
This is a follow up to #341.  Now that the code has an implementation for the RIM event digest this now needs to be included in the firmware validation.

PR #352, opened 2021-04-15
[#343] Event digest update (UI)

cyrus-dev:
In an effort to keep the PR a bit smaller.  This is the first part of event digest validation that is associated with #343.  

The first main thing this PR addresses is updating the UI to display the differences in support RIMs whether they are Supplemental or Patch. 

Because of updating the UI, more had to be updated in the background to make sure the support rims were updated with the correct information from the associated base, which depends on when each are uploaded and how they associate with one another.  A lot of the code change have to do with moving code around so that different fields within the RIM objects could be updated.

Also fixed a visual error that causes all support RIM events to show up red (error).  This had to do with the comparing of events by event number, which will never match up, to using the PCR Index.  This does part of the validation but only for page display, not during the Firmware Validation process.


iadgovuser26 (approved):
:+1:

PR #353, opened 2021-04-21
CA for the Base RIM validation

cyrus-dev:
Updated code to fail validation if the RIM signing certificate fails to validate against its own issuer.


cyrus-dev:
This branch conflicts with another issue.

PR #361, opened 2021-05-24
[#359] Check base64 encoding for base RIM attributes

### LINKED ISSUE #359: Check base RIM attribute values for base64 encoding
The current tcg_rim_tool encodes neither the RimLinkHash value read from the JSON attribute file nor the file hash calculated from the support rim.  Furthermore, the tool does not check the RimLinkHash value read from the JSON file.

Changes related to this ticket will:
1. Check the RimLinkHash value read from the JSON attributes file.  If it is already in base64 encoding it will be written directly to the output base rim.  If it is not, it will be encoded, the user will be alerted via commandline message, and the resulting encoded string will be written to the output base rim.
2. Encode the calculated payload file hash in base64 before writing to the output base rim.

chubtub:
Encode the calculated Payload file hash in base64 before writing out to the base RIM.

Closes #359

### DIFF
--- a/tools/tcg_rim_tool/src/main/java/hirs/swid/utils/HashSwid.java
+++ b/tools/tcg_rim_tool/src/main/java/hirs/swid/utils/HashSwid.java
@@ -23,7 +23,7 @@ public class HashSwid {
 
     /**
      * Getter method for the hash that uses 256 bit hash
-     * @param value
+     * @param filepath the file to hash.
      * @return 
      */
     public static String get256Hash(String filepath) {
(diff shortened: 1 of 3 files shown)


chubtub:
After reviewing the SHA256 spec (`http://www.w3.org/2001/04/xmlenc#sha256`) it was determined that the rimLinkHash value itself was not necessarily to be in base64 encoding, but rather the value in <DigestValue>.  The changes in this PR have been updated to remove rimLinkHash encoding.

iadgovuser26 (approved):

PR #372, opened 2021-06-10
Base64 decode exception

cyrus-dev:
Added an additional catch statement to the parsing of certificates that are PEM in case of a DecoderException for Base64.decode method.  Instead of going to a blank page with the error, the ACA catches the exception and states on the page in which the file was uploaded to.


chubtub (approved):
Changes look good

iadgovuser26 (approved):

PR #387, opened 2021-08-17
[#386] Update overloaded validateCertChain to check cert chain consistently

### LINKED ISSUE #386: Update HIRS_Utils.validateCertChain method
HIRS_Utils.validateCertChain is an overloaded method that checks a given X509Certificate or X509AttributeCertificateHolder against a collection of other certs for a complete validation chain.  The method that accepts an X509Certificate parameter was updated in #379, the method that accepts a X509AttributeCertificateHolder parameter should be updated similarly so that the order of objects in the collection of other certs is not a point of failiure.

chubtub:
Refactor to account for any order of objects in additionalCerts parameter. 

Closes #386

### DIFF
--- a/HIRS_Utils/src/main/java/hirs/validation/SupplyChainCredentialValidator.java
+++ b/HIRS_Utils/src/main/java/hirs/validation/SupplyChainCredentialValidator.java
@@ -40,7 +40,6 @@ import java.security.NoSuchProviderException;
 import java.security.PublicKey;
 import java.security.Security;
 import java.security.SignatureException;
-import java.security.cert.CertificateEncodingException;
 import java.security.cert.CertificateException;
 import java.security.cert.CertificateExpiredException;
 import java.security.cert.CertificateNotYetValidException;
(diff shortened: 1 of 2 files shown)


cyrus-dev (approved):

HIRS_Utils/src/main/java/hirs/validation/SupplyChainCredentialValidator.java
```diff
@@ -1338,43 +1339,49 @@ public static boolean verifyCertificate(final X509Certificate cert,
...
+                                                                                    trustedCert);
+                if (issuerMatchesSubject && signatureMatchesPublicKey) {
+                    if (isSelfSigned(trustedCert)) {
+                        LOGGER.info("CA Root found.");
+                        return "";
+                    } else {
+                        foundRootOfCertChain = intCAError;
+                        try {
+                            startOfChain = new X509AttributeCertificateHolder(
+                                                            trustedCert.getEncoded());
```
iadgovuser29, line 1365: trustedCert here is a key certificate, not an attribute cert. This line is causing the following error: "Error checking cert chain: malformed data: unknown object in getInstance: org.bouncycastle.asn1.ASN1Integer"

iadgovuser29: The commit below should have fixed this problem.

iadgovuser29 (approved):

cyrus-dev (approved):

iadgovuser26 (approved):

PR #407, opened 2021-10-21
[#381] Update RIM validation in ACA

### LINKED ISSUE #381: ReferenceManifestValidator public key parsing
As was discovered in the course of completing #375 the public key parsing in ReferenceManifestValidator needs to be updated.  The changes should be similar to those in tcg_rim_tool's SwidTagValidator class.

chubtub:
- Modify ACA RIM validation to search for a signing cert if the base RIM does not have an embedded cert.
- Validate the ca chain of the found signing cert.

The changes to the RIM controller and SupplyChainValidationServiceImpl are summarized by the following:
1. Parse from the RIM the embedded cert if it is present, or the subject key identifier (skid).
2. Loop through all certs found and compare to the embedded cert or skid for a match.
3. If a match is found, validate the Signature element of the RIM.
4. If the Signature element is valid, get the signing cert's respective CA chain and attempt to validate.

Closes #381

### DIFF
--- a/HIRS_AttestationCA/src/main/java/hirs/attestationca/AbstractAttestationCertificateAuthority.java
+++ b/HIRS_AttestationCA/src/main/java/hirs/attestationca/AbstractAttestationCertificateAuthority.java
@@ -458,7 +458,7 @@ public abstract class AbstractAttestationCertificateAuthority
             // Package response
             ProvisionerTpm2.IdentityClaimResponse response
                     = ProvisionerTpm2.IdentityClaimResponse.newBuilder()
-                    .setCredentialBlob(blobStr).setMask(pcrQuoteMask)
+                    .setCredentialBlob(blobStr).setPcrMask(pcrQuoteMask)
                     .build();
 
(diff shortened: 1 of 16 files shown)


iadgovuser29:
At some point, we need to work on the server responses and codes so that the client can provide better feedback in fail conditions.

For now, if quote validation fails on the ACA, the provisioner will end with:
```
----> Provisioning the quote failed.Please refer to the Attestation CA for details.
```

If the ACA stops due to some validation problem prior to creating the MakeCredential blob, for example if the certification path of an artifact could not be validated, the provisioner will end with:
```
[libprotobuf ERROR google/protobuf/message_lite.cc:123] Can't parse message of type "hirs.pb.IdentityClaimResponse" because it is missing required fields: credential_blob
----> Provisioning failed.
The ACA did not send make credential information.
```

chubtub (commented):
Changes look good

chubtub (commented):
Changes look good.

PR #416, opened 2021-11-09
Vendor Table JSON refactor

cyrus-dev:
This commit updates the location of the vendor-table.json which is used by the ACA and the eventlog tool.  The previous change moved the vendor-table.json to a separate folder, however the code was not updated to reflect the change.  This instead updates how the two install handle the file.  The ACA was updated not to clobber the file if it exists and the eventlog tool was updated to do the same through an additional script during post install.

This was tested by installing the ACA first, then the eventlog tool.  And then tested in reverse, both attempts passed without crashing.


chubtub (approved):
Changes look good

iadgovuser26 (approved):

PR #420, opened 2021-11-10
RimLinkHash Fix

cyrus-dev:
The rimlinkhash meta information wasn't linking up with the associated swidtag.  This is because the wrong hash look up was being used.  Previously when the hexDecHash and base64Hash were implemented, the main focus was on the rimel and not the swidtag.


chubtub (approved):
Changes look good.

PR #446, opened 2022-02-23
[ #445] Added Initial Firmware Validation System Tests

### LINKED ISSUE #445: Need to add FW Validation Tests using PC Client RIMs
The System Tests need to exercise Firmware Validation using PC Client RIMs. 
The test should include a Primary Base RIM (from an OEM) and Supplemental (from a VAR).

iadgovuser26:
Adds 3 CI test to cover 
1 good test from 2 sources a OEM and VAR.
2 failure cases:

-  a Primary RIM with an event that does not match the TCG event log

-  a Supplemental RIM with an event that does not match the TCG event log 

closes #445


cyrus-dev (approved):
I tried eyeballing them but as long as the names match up between swidtag and rimel, looks good.

PR #488, opened 2022-08-30
[#487] Create xml_dsig_tool

### LINKED ISSUE #487: Create an alternative tool for validating rim signatures
There are a limited set of open source or online tools that can validate xml based signatures based upon [XML Signature Syntax and Processing Version 2.0](https://www.w3.org/TR/xmldsig-core2/) , which is required by the [PC Client RIM specification](https://trustedcomputinggroup.org/wp-content/uploads/TCG_PC_Client_RIM_r1p04_pub.pdf). CSharp provides an xml library that should be able to create and verify xml signatures  which could serve as a secondary check on the xml signature applied to RIMs. This issue will investigate the creation of an xml_dsig_tool based upon csharp.

chubtub:
This PR creates a C# implementation of the xml-dsig spec, with sign and validate functions.  The intent of this tool is to sign and validate xml documents (e.g. base RIMs) interchangeably with the tcg_rim_tool.

There is a debug function that is a work-in-progress.  The original goal was to read a signed or unsigned xml document and print some information regarding its expected signature in such a way as to assist with errors encountered in signing or validating the document.

Closes #487


iadgovuser26:
Need to add a README.md

iadgovuser26 (approved):

PR #494, opened 2022-09-21
[#493] Support RFC3339 format timestamp

chubtub:
These changes add support for timestamping XML signatures of base RIMs according to RFC3339.  The timestamp object is inserted into the signature via an XMLObject by way of the commandline argument `--rfc3339`.

This PR addresses #493 together with #505.

Issue #493, opened 2022-09-16
Support RFC3852 and RFC3339 for timestamps


chubtub:
Add command line options to distinguish between timestamps that comply with RFC 3852 and RFC3339.

PR #505, opened 2022-12-08
[#493] RFC3852 counter signature timestamp

chubtub:
These changes provide support for timestamping base RIMs with RC3852-compliant counter signature data.  A flat file containing the counter signature is passed in with the `--rfc3852 <file>` command line option, and its base64 encoding is placed in the Timestamp element of the resulting RIM.

This PR addresses #493 together with #494.


chubtub:
Superseded by #506

PR #509, opened 2023-03-13
[#497] Support for composite RIMs

### LINKED ISSUE #497: Composite RIM support
Provide support for composite RIMs, wherein a single base RIM references one or more components in its payload.  Each component RIM should be referenced by a hash of itself and its file location.  Component RIMs may be signed by entities that are not the signer of the referencing base RIM, so validation would include gathering all relevant certificate chains.

chubtub:
Modified rimtool to support multiple payload files for creating and validating base RIMs.
- Gateway class expects a JsonArray for the File object in the attributes file
- Validator class validates all File elements under Directory against their declared hashes
- `--directory <path>` causes the rimtool to search <path> for the files indicated in the attributes file

Closes #497


chubtub:
Superseded by #527

Issue #531, opened 2023-06-13
gitHash not working in Eclipse


iadgovuser59:
In Eclipse 2023-03, `gitHash` in `build.gradle` causes the Gradle build to fail, likely due to an issue with the working directory when executing the process.

PR #533, opened 2023-06-14
Fix to build.gradle on Eclipse

iadgovuser59:
Converts command in `build.gradle` to use `ByteArrayOutputStream` in order for it the git command to work properly in Eclipse.

Fixes issue #531.


cyrus-dev:
Tested fix and the issue is resolved.

Issue #534, opened 2023-06-15
UI Migration Clean up


cyrus-dev:
Now that the data can upload (RIM and Certificate), showing it in "/list" is generating display errors.  

1. **All certificates are returned, not the specific type** (observe uploading a platform certificate and then note it shows up under all certificate list pages)
2. **All items are shown instead of the set default "10"** (observe by uploading base/support rim and navigating to the events display page and not all events show up on the page instead of just 10)
3. Downloading items
4. Base RIM details page coming up empty
5. Base RIM details page "Files" collapse and uncollapse function not working due to #529 issue that should resolve it
6. Upload allows for multiple items
7. Support RIM deletion leaves associated rim events in the db and if the same file it re-upload, the duplicates are uploaded as well

cyrus-dev:
The main issues have been taken care of

PR #542, opened 2023-06-20
[#534] Fix for Downloading Certificates

cyrus-dev:
Added download method mapper for certificates

#534 item 3 resolution


iadgovuser29 (approved):
I can see the button to download an individual certificate works now.

Issue #517, opened 2023-05-11
 RIM mismatch when RIM files are uploaded via the RIM page


iadgovuser26:
When uploading the Support RIM and Support RIM vs having it uploaded from the provisioner application the ACA may have a problem associating the RIM files with the provisioning request. The use of the platformManufacturerStr and platformModel from the Base RIM (swidtag) will be used to match the against provisioning request.

PR #543, opened 2023-06-21
[#534] Updated display code

cyrus-dev:
This check in updates the code to list the certificates on their specific page, rather than all on any of them.  Next will be introducing the new parsing code for RIM and TLS so that Trust Chains show up

This covers 1. on the issue #534 

The amount of changes was growing so this is a stable push.  Next will be to resolve the displaying of RIMs and that is a part of #517 and next implementing the TLS for Trust Chain list displaying.


iadgovuser29 (approved):
Upload of key certificates works. Certificates show up in their expected tables.

Issue #536, opened 2023-06-15
Create filter for Github actions to trigger


iadgovuser26:
Currently several Github actions run on all pushes of code. The code being assembled for the migration effort is triggering these actions without having the tests ready.  The proposed fix is to add a branch filter to only trigger the actions based upon branch name containing a "v2" in the name. This would effect the following actions:
* HIRS Build and Unit Test
* HIRS packages for centos
* HIRS System Tests

Any code intended for the master branch should use a branch name with a  lower case "v2" in the name.
The "main" branch will be used as the target branch for the migration effort and not currently trigger these actions.
Any code intended for the main branch should use a branch name with a lower case "v3" in the name.

iadgovuser26:
Closed by PR 545

PR #544, opened 2023-06-23
[#534] Base RIM Details Page

cyrus-dev:
the DN compare of bouncy castle utils throws and exception if the object is null.  Added additional condition to the if statement to check null status.

Screenshot at 2023-06-23 09-09-26

#534


cyrus-dev:
This last push resolves the issue #536

iadgovuser29 (approved):

PR #552, opened 2023-07-11
Reference Digest value UI Update

cyrus-dev:
This PR fixes issues with displaying associated measurement events for the Reference Manifest support file.  If the support file is deleted, it now removes all associated events.  Also the repository calls are updated and the main referencing will be done by file hash instead of UUID.

This PR fixes issues associated with item 7 in #534


iadgovuser29 (approved):
Verified that records are now removed from the database when a support RIM is deleted. Also verified it doesn't delete extra records from other RIMs uploaded to the same database.

Issue #546, opened 2023-06-29
Create unit test for TCGEventLog.java under ..utils/tpm/eventlog in main


iadgovuser58:
branch:

iadgovuser26:
closed by #557

PR #556, opened 2023-07-19
[#546] Add TCGEventLogTest

iadgovuser58:
Adds unit test TCGEventLogTest and 4 test files


iadgovuser58:
Need to make a few more changes

PR #584, opened 2023-09-14
Initial Setup for ACA provisioning

cyrus-dev:
This PR push the starter entry point code back in for the ACA to receive requests from a client.  This code doesn't include the main meat of provisioning and determining a failure.

Some specific noted changes, I parsed out the original file that was the entry point as it was one very large file.  I broke it out based on the 3 types of requests that come in.  Added in a helper method to print out the expect URIs for Restful processing.

I tested this with an old cpp provisioner, there are updates in this for that provisioner for testing purposes.


iadgovuser29 (approved):
These are initial changes to the ACA to enable REST POST requests. As mentioned in the PR, more work is needed to get provisioning and validation set up in v3.

PR #605, opened 2023-10-13
HIRS ACA Provisioning

cyrus-dev:
This set of changes includes all remaining code to provision a client.  This will likely not be the last bit however with the amount of changes it is ready to be merged and is functioning.  

Remaining elements that'll need to be exercised is quote validation and firmware.  These will be revisited when we have more testing environments.  I have successfully passed the Endorsement and Platform policies.  Issued certificates are being created and validation reports are populating and linking.  

There is still question about the Trust store download, if that is even needed.

I have resolved the majority of bugs that I have encountered while testing.  I used 3 separate devices to test.


iadgovuser29 (approved):
EK and Platform Certificate signature validation works as described. Platform Attribute validation does not yet work.

PR #609, opened 2023-10-16
Creates a build/package action for V3

iadgovuser26:
Adds an action that creates and archives all HIRS jar files, a  bootable war file, an ACA RPM and ACA DEB package for each push.

Closes #606


iadgovuser58 (approved):
Approved

PR #620, opened 2023-10-31
Update to the RIM Validator

cyrus-dev:
Updated the Rim Validator to remove direct object parameters from the database and pass in the information the methods needed to function


chubtub (approved):
Changes look good

PR #630, opened 2023-11-14
V3 server UI fixes

cyrus-dev:
This pull request has some bug fixes and migrated functionality that was not included in the original push.  Due to unit test migration, there was a problem found with a unit test that calls the archive method for DB entities.  This function wasn't originally set up to work.  This PR includes the call and fixes up issues with the /list portion of the ACA portal displaying the correct number of items displayed and not the number in the DB.

In addition, while testing there was a mistake that led to a 500 error that wasn't covered in the try/catch handling.  When uploading a rim to a certificate page and vice versa; these actions produced 500 errors for IllegalStateException and Nullpointer.   The nullpointer can't be handled by our code because it is in the parsing libraries that are imported.

This does not resolve the issue listed in #619.


iadgovuser58 (approved):
Looks good to me. This is minor but I notice sometimes "this." isn't consistently used like ArchivableEntity lines 63-66 and a few other places.

Issue #642, opened 2023-11-29
Add spot-bugs to Grade build


iadgovuser26:
[Spot bugs ](https://spotbugs.readthedocs.io/en/latest/gradle.html) replaces the findbugs tool used in previous builds.  This tool should be made available via a gradle task.

cyrus-dev:
Splitting this into multiple PRs to not have a massive push.  First part just included the small set of changes in eventlog.  Next is AttestationCA that listed 202 errors.

PR #653, opened 2023-12-20
[#642] Initial Spotbugs Setup

cyrus-dev:
This is the first step for adding spotbugs (formerly findbugs) back into the build.  The configuration is in place but the majority of the projects are being excluded.  The eventlog (a small sample size) is however being run with the proper corrections committed.

This is the first part of #642 

To initial the "spotted" bugs you have to take out the line in the exclude file, the entire match tag.
`<Match>`
`        <Package name="~hirs\.structs.*" />  `
`   </Match>`

as an example for HIRS_Structs


chubtub (approved):
Changes look good

PR #652, opened 2023-12-15
Additional Updates/Fixes for UI and Unit Tests (wip)

cyrus-dev:
This pull request includes updates to:

- Bouncy Castle, updating the version from 1.70 (2021) to 1.77 (2023)
- An adjustment to the creation of the create time stamp for archivable objects.  They were not updating when the reset method was called. 
- Updating the the EK link not showing up on the platform certificate /list page

PR #654, opened 2023-12-20
Creation Date fix

cyrus-dev:
These are fixes for issues found via unit test that are in PR #652 .  But that PR is currently still WIP and these changes have nothing to do with why that PR is WIP.  Therefore PR just covers these changes to the creation time not being saved/reset.


iadgovuser58 (approved):
Works great

PR #671, opened 2024-01-16
[#668] Added VERSION file to /opt/hirs

iadgovuser26:
Added Version file to:

- /opt/hirs/aca when installing the ACA from rpm or deb 
- /opt/rimtool when installing the tcg_rim_tool  from rpm or deb 
- /opt/eltool when installing the tcg_eventLog_tool from rpm or deb  

Other minor updates

- tcg_rim_tool updated to use gradle 8.3
-  tcg_rim_tool installs to /opt/rimtool
- tcg_eventLog_tool updated to use gradle 8.3
- tcg_eventLog_toolinstalls to /opt/eltool

closes: #668

PR #676, opened 2024-01-17
Verison helper update

cyrus-dev:
The version number was not showing up on the ACA Portal.  This was due to the changes in how the ACA was packaged and loaded when switching from v2 to v3.  The location has been updated in #671, this PR updates the code to use the new location.
image


chubtub (approved):
Changes look good

Issue #651, opened 2023-12-14
Remove @EqualsAndHashCode annotation for JPA entities


iadgovuser59:
Some JPA entities currently use the Project Lombok `@EqualsAndHashCode` annotation to auto-generate `equals()` and `hashCode()` overrides respectively. However, there are some pitfalls in doing so, some of which were encountered during unit test conversion.

It would seem that some Java community discussions ([example](https://stackoverflow.com/a/5729992), among others) recommend manually overriding `equals()` and `hashCode()` so as to make behavior more predictable when comparing objects or using JPA lifecycle management. Based on this, it may be possible to run into these issues once two entities are compared, leading to unexpected behavior.

This ticket is to investigate whether removal of these annotations from JPA entities in HIRS is necessary, and if so, to remove them and replace them with `equals()` and `hashCode()` implementations as found in `master`. Afterwards, we may need to copy over the corresponding portions of unit tests from `master` for these methods to ensure consistency.

PR #679, opened 2024-01-19
[#642] HIRS_Utils spotbugs fixes

cyrus-dev:
Initial set of changes for #642 that are fixes for HIRS_Utils. In addition, this and the following pushes will have changes for #651 because spot bugs doesn't like how Lombok handles hashCode and equals


iadgovuser26 (approved):

chubtub (approved):
Changes look good

PR #683, opened 2024-01-24
Bouncy Castle version update

cyrus-dev:
Bouncy Castle, updating the version from 1.70 (2021) to 1.77 (2023)

While expecting a StackOverFlow error from uploading an EK certificate,
this is not occurring.  A unit test had to be updated because the
previous BouncyCastle X500Compare.equals() method returned false for 2
empty string compares.  Now it returns true.


chubtub (approved):
Changes look good

PR #691, opened 2024-01-25
[#661] Output handling should be more concise

### LINKED ISSUE #661: Output options for rimtool operation
The rimtool outputs the same list of input arguments for both creation and validation, regardless of the use case.  Modify the output to be less verbose unless explicitly requested.

chubtub:
Superseded by #715 

These changes include the following:

1. This block of text is printed following `-c` and `-v` functions only if `--verbose` is also given

`Creating: base`
`Using attributes file: HIRS-issue_103-test-rim-configs/configs/Base_Rim_Config.json`
`Write to: `
`Verify file: `
`Truststore file: default (/opt/rimtool/data/keystore.jks)`
`Event log support RIM: HIRS-issue_103-test-rim-configs/eventlogs/TpmLog.bin`
`No timestamp included`

2. Unrecognized command line options will cause the rimtool to exit, and all unrecognized options will be printed.

Close #661


chubtub:
Superseded by #715

PR #700, opened 2024-02-01
[#692] WIP: validate rimtool arguments

### LINKED ISSUE #692: Rimtool credential checking
The rim tool help defines -p as "-p, --publicCertificate <path> The public key certificate to embed in the base RIM created by this tool. " In this case the -p was used for embedding the cert as well as a verify. It's a question of if the verify should accept the -p option.  

Seems like the verify currently does accept the -p option but it didn't check the entire cert chain with the -p option. If that's by design we need to spell that out very clearly and expect a good amount of confusion when it gets used out there. If that wasn't the intent it sounds like it needs to check the entire chain and fail. 

The PC Client RIM does state that the embedded cert cannot be a self signed cert so a trust store is required for a verify. In that context a -p without a -t for a verify of a base rim with an embedded cert should realistically never pass. Perhaps -p for a verify should check for the full chain and abort if it doesn't find one or just fail if a -p is encountered or a -t is not present.  An error statement  pointing out that the -p is not valid for the command requires a -t option would be helpful. 

Another item I noticed the help description for -t reads:
-t, --truststore <path>
      The truststore to sign the base RIM created or to validate the signed 
      base RIM.

The key is used to sign. Trustore is just used to verify. It should probably read "The truststore used to validate the signed  base RIM"

chubtub:
These changes will check that the proper credential arguments are given, and log appropriate errors if not.

Closes #692


chubtub:
This first push clarifies an exception that is thrown when the signing cert for a signed RIM is not found in the provided truststore.

chubtub:
These changes have been merged into #715.

Issue #705, opened 2024-02-06
Investigating ACA UI Functionality


cyrus-dev:
There are several issues found while using the new set of code that are older or new.  This is to track the remaining:

1. The search function of the datatable of the /list pages
2.  Like the search function, the column sort function is working but not as intended for the /list
3. The validation page green check mark is showing doubled up in certain columns
4. Updating the policy page with a newer policy to ignore the system bios update component from the provisioning of the platform certificate attributes

PR #706, opened 2024-02-06
[#705] PC Attributes Provision Error

cyrus-dev:
This commit is for the 3rd listed issue on #705.  The green check mark test was causing an error from ASN1UTF8String.getInstance.  And did some checkstyle fixes.

To replicate, you should just need to select PC attribute policy and provision.  My test was done on a dell.


iadgovuser29:
[This](https://github.com/nsacyber/HIRS/commit/f7103517d21a0a89a145de3ad0c50b80b70ed233#diff-d3e11a5a87064d976addc9450625efba46d838770d05a6e5fd1296606ff6aa8f) was the commit that changed ```DERUTF8String.getInstance``` to use ```ASN1UTF8String.getInstance```. 

I know that's the recommendation in the Javadoc for the deprecated DERUTF8String version. But ASN1UTF8String can't seem to be able to handle Strings.

Use ```new DERUTF8String()``` instead.  There are a few more usages of the ASNUTF8String version of getInstance that might cause provisioning issues.

iadgovuser26 (approved):

Issue #723, opened 2024-02-28
ACA fails on Platform Version check


iadgovuser26:
**Problem**: The "Platform Version" field as described in section 2.1.5.6 of the v1.1 version of the [ TCG platform cert spec ](https://trustedcomputinggroup.org/wp-content/uploads/IWG_Platform_Certificate_Profile_v1p1_r19_pub_fixed.pdf)  is listed as a MUST in the table of section 2.1.5 so a Verifier is obligated to check/process it. It does have the following note: "platform version is encoded as a string and is the manufacturer-specific implementation version of the platform", so a manufacturer could technically add whatever value it wanted.  Any value, however, makes it rather difficult for Verifier to match in this case.

The Platform Version attribute within a Platform Certificate typically defaults to either "Unknown" or "Not Specified" if the platform does not track its version (many devices will typically change a model number versus tracking a version). This can lead to a mis-compare by the ACA during the provisioning process (i.e. a false negative).
 
**Proposed solution**:   If the ACA finds a Platform Version field within a Platform certificate with a value of  either "Unknown" or "Not Specified" (not case sensitive)  it should skip the comparison with the actual value returned from the provisioner. If it is not one of the the two values it should proceed with the comparison.

PR #725, opened 2024-02-29
[#723] Update Provisioning for Version value

cyrus-dev:
Updating the code to ignore 'Not Specified' and 'Unknown' as values for System Information's Platform Version.


cyrus-dev:
This now includes changes that will start satisfying #705 , the 5th issue on that issue.

iadgovuser29 (approved):

PR #730, opened 2024-03-07
[#705] Component failure highlight part 2

cyrus-dev:
This is a continuation of the PR for 723 and issue 705.  These changes set up fixing the delta part of the provisioning of the attributes.

This is the second part which covers the delta portion of this code.


cyrus-dev:
PR 733 covers this request now

Issue #707, opened 2024-02-08
Add ACA policy to ignore component revision within the Platform Cerificate


iadgovuser26:
Within the ComponentIdentifier of the Platform Certificates is a componentRevision field. While an exact match should be required for most Supply Chain scenarios there are use cases in which the Platform Certificate may be used for system monitoring use cases which must accommodate for component firmware updates. Component revisions (e.g. System BIOS revision) get systematically updated and a verification of the component will currently fail. 

Proposed ACA Policy addition:
* Platform Credential Validation:
      * Ignore component revisions: Disabled

Default should be set to Disabled

cyrus-dev:
This is also reference in #705

iadgovuser26:
Closed by PR !733

PR #732, opened 2024-03-08
[#707] Add Policy to Ignore Component Revision

cyrus-dev:
image

Added UI features and back end code to handle the new policy which doesn't include the revision comparison as part of the final validation pass/fail.


cyrus-dev:
PR 733 covers this request now

PR #766, opened 2024-04-26
updated build to include HIRS_UTILS

iadgovuser26:
Creates a "Fat jar" for running without linking in HIRS_Utils.jar for the tcg_eventlog_tool.


chubtub (approved):
Changes look good, the assembled fat jar contains HIRS_Utils classes.

PR #786, opened 2024-06-13
Changed DB lookup of RIM EL objects, added null checks to Device.toString

### LINKED ISSUE #777: NullPointerException in Device.toString
https://github.com/nsacyber/HIRS/blob/4b72d9aee39c6072a807080bfe6251560933d71e/HIRS_AttestationCA/src/main/java/hirs/attestationca/persist/entity/userdefined/Device.java#L115-L118

The value of this toString function is questionable. Make sure the information returned in this function is necessary. For instance, the portal may use this string directly. In that case, consider using a different function name.

toString is also used by IDEs when debugging. There is not enough null checking in this code at the moment. In particular, healthStatus is often null and will cause the object to look broken during debugging because toString makes a null reference.

The change to use the spring @repository annotation obscures how spring retrieves information from the database. And a null pointer exception in this toString method can make debugging very difficult. DeviceRepository's findDeviceByName function does retrieve the correct object. The IDE will show that null pointer exception unless you look closely at the object.

I don't think it hinders validation. It's not an urgent problem yet.

### LINKED ISSUE #783: Potential issue with Quote Validation
TPM quote validation is halting early on this line https://github.com/nsacyber/HIRS/blob/main/HIRS_AttestationCA/src/main/java/hirs/attestationca/persist/service/SupplyChainValidationService.java#L362
```SupplyChainValidationSummary previous = this.supplyChainValidationSummaryRepository.findByDevice(deviceName);```

It's not clear yet why that lookup fails. It could be because deviceName here is a string, and the only reference to a Device in SupplyChainValidationSummary is by object. Something need to look in to.

The CI test should probably not be passing. Validation might be returning a success even with the quote validation stopping.

### PR DESCRIPTION
**Pull request by iadgovuser29**
Closes #777, closes #783 

The CI tests now complete properly. The issued attestation certificate is successfully delivered to the client.

### DIFF
--- a/HIRS_AttestationCA/src/main/java/hirs/attestationca/persist/entity/userdefined/Device.java
+++ b/HIRS_AttestationCA/src/main/java/hirs/attestationca/persist/entity/userdefined/Device.java
@@ -114,8 +114,8 @@ public class Device extends AbstractEntity {
 
     public String toString() {
         return String.format("Device Name: %s%nStatus: %s%nSummary: %s%n",
-                name, healthStatus.getStatus(),
-                supplyChainValidationStatus.toString());
+                name, (healthStatus == null ? "N/A" : healthStatus.getStatus()),
+                (supplyChainValidationStatus == null ? "N/A" : supplyChainValidationStatus.toString()));
     }
 
     @Override
--- a/HIRS_AttestationCA/src/main/java/hirs/attestationca/persist/service/SupplyChainValidationService.java
+++ b/HIRS_AttestationCA/src/main/java/hirs/attestationca/persist/service/SupplyChainValidationService.java
@@ -321,7 +321,7 @@ public class SupplyChainValidationService {
                                     deviceName));
                 } else {
                     ReferenceManifest manifest = referenceManifestRepository
-                            .findByHexDecHashAndRimType(sRim.getEventLogHash(), sRim.getRimType());
+                            .findByHexDecHashAndRimType(sRim.getEventLogHash(), ReferenceManifest.MEASUREMENT_RIM);
                     if (manifest instanceof EventLogMeasurements) {
                         eventLog = (EventLogMeasurements)manifest;
                     }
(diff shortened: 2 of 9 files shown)

### CODE REVIEW

**Review by iadgovuser26 — APPROVED** [review]

File: HIRS_AttestationCA/src/main/java/hirs/attestationca/persist/entity/userdefined/Device.java [hunk 1]
```diff
@@ -114,8 +114,8 @@ public void setLastReportTimestamp(final Timestamp lastReportTimestamp) {
 
     public String toString() {
         return String.format("Device Name: %s%nStatus: %s%nSummary: %s%n",
-                name, healthStatus.getStatus(),
-                supplyChainValidationStatus.toString());
+                name, (healthStatus == null ? "N/A" : healthStatus.getStatus()),
```
chubtub, line 117: Should this be HealthStatus.UNKNOWN instead of "N/A"?

HIRS_AttestationCA/src/main/java/hirs/attestationca/persist/entity/userdefined/Device.java
```diff
@@ -114,8 +114,8 @@ public void setLastReportTimestamp(final Timestamp lastReportTimestamp) {
...
+                (supplyChainValidationStatus == null ? "N/A" : supplyChainValidationStatus.toString()));
```
chubtub, line 118: AppraisalStatus.UNKNOWN?

HIRS_AttestationCA/src/main/java/hirs/attestationca/persist/service/SupplyChainValidationService.java
```diff
@@ -11,10 +11,7 @@
 import hirs.attestationca.persist.entity.manager.ReferenceManifestRepository;
 import hirs.attestationca.persist.entity.manager.SupplyChainValidationRepository;
 import hirs.attestationca.persist.entity.manager.SupplyChainValidationSummaryRepository;
-import hirs.attestationca.persist.entity.userdefined.Device;
-import hirs.attestationca.persist.entity.userdefined.PolicySettings;
-import hirs.attestationca.persist.entity.userdefined.SupplyChainValidation;
-import hirs.attestationca.persist.entity.userdefined.SupplyChainValidationSummary;
+import hirs.attestationca.persist.entity.userdefined.*;
```
chubtub, line 14: Checkstyles will flag this as an import group to expand.

chubtub (commented):
I left a few comments, otherwise changes look good.

chubtub (approved):
Changes look good

PR #796, opened 2024-07-01
Move part of PciIds from AttestationCA back to Utils

### LINKED ISSUE #757: Move PciIds from AttestationCA back to Utils
The class PciIds is in AttestationCA. However, PciIds is needed to process SPDM event log structures defined in the PFP (specifically DEVICE_SECURITY_EVENT_DATA_PCI_CONTEXT), and the event log classes are in Utils. PciIds class cannot be accessed from Utils without creating a circular dependency. Need to 1) move PciIds back into Utils, 2) Update DeviceSecurityEventDataPciContext to use PciIds to report Vendor and Device IDs.

PciIds depends on 3 classes in AttestationCA, which in turn depend on more classes. To move PciIds, ALL of these classes will have to move to Utils. Here are the following dependencies:

PciIds depends on:
- ComponentIdentifier
- ComponentIdentifierV2, which depends on:
  - ComponentAddress
  - ComponentClass
  - ComponentIdentifier [see above]
  - URIReference
- ComponentResult, which depends on:
  - ArchivableEntity, which depends on:
    - AbstractEntity
  - ComponentAddress [see above]
  - ComponentIdentifier [see above]
  - AttributeStatus
  - ComponentIdentifierV2 [see above]

iadgovuser58:
Closes #757 

The class PciIds is in AttestationCA. However, a few functions in PciIds are needed to process SPDM event log structures defined in the PFP (specifically DEVICE_SECURITY_EVENT_DATA_PCI_CONTEXT), and the event log classes are in Utils. PciIds class cannot be accessed from Utils without creating a circular dependency. 
1) renamed AttestationCA pciids to AcaPciIds
2) created a Utils PciIds and moved the DB object, the vendor id and device id translate functions in AcaPciIds into Utils PciIds
3) updated DeviceSecurityEventDataPciContext to use PciIds to report Vendor and Device IDs
4) update Utils PciIds to also translate classcode, and update DeviceSecurityEventDataPciContext to use this feature

### DIFF
--- a/HIRS_AttestationCA/src/main/java/hirs/attestationca/persist/util/AcaPciIds.java
+++ b/HIRS_AttestationCA/src/main/java/hirs/attestationca/persist/util/AcaPciIds.java
@@ -14,7 +14,8 @@ import org.bouncycastle.asn1.DERUTF8String;
 import java.util.ArrayList;
 import java.util.List;
 
-import static hirs.utils.PciIds.DB;
+import static hirs.utils.PciIds.translateDevice;
+import static hirs.utils.PciIds.translateVendor;
 
 /**
  * Provide Java access to PCI IDs.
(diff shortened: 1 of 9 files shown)


iadgovuser58:
This was build off PR #795 which hasn't been merged yet, so it looks like more files changed than it actually is (only the last few commits belong to this PR).

iadgovuser29 (approved):

PR #805, opened 2024-07-11
supplychainvalidationsummary will look up rims by ID

iadgovuser29:
Addresses a base rim look up issue with the supplychainvalidationsummary repository

Some CI tests weren't working because the TPM in our container was not being reset between tests. Any test that extended a PCR would affect subsequent tests.

Moved several TPM functions to a common script. Those functions will currently only work under certain conditions:
* They assume the IBM sw tpm server repo is cloned to /ibmswtpm2 and is built.
* They assume the IBM tss repo is cloned to /ibmtss and is built.
* They assume tpm2-tools are installed.
* They assume the HIRS repo is cloned to /hirs.


iadgovuser29:
Changed look up of RIM during validation in the case that a person uploads the RIM separately to the portal.

Added a CI test for this case.

iadgovuser26 (approved):

Issue #740, opened 2024-03-21
Update .NET Provisioner to handle an LDevID Certificate returned at the end of provisioning


iadgovuser29:
The .NET Provisioner can store the LDevID Certificate created and delivered by the ACA in a similar way as the AK certificate.

Extend the protobuf file to allow the LDevID certificate to be passed at the end of provisioning.

Issue #739, opened 2024-03-21
Pass LDevID Public Key to the ACA


iadgovuser29:
Relies on #738

Retrieve the public key created by #738 and include it in the information passed to the ACA.

The protobuf file will need to be extended to have a place for the key.

Issue #738, opened 2024-03-21
Create LDevID Key using .NET Provisioner


iadgovuser29:
The .NET Provisioner needs to create the LDevID key pair.

The object attributes for the LDevID must include: fixedTPM, sensitiveDataOrigin, sign.
The object attributes for the LDevID must be NOT Restricted and NOT decrypt.

The LDevID probably should be created by the provisioner at the same time as the AK.

Issue #709, opened 2024-02-09
Implement LDevID issuing logic


iadgovuser59:
The goal of this task will be to implement the logic to issue an LDevID certificate for the DevID class created in an earlier task. Broadly speaking, this will involve the following steps:

1. Work with a CA trust chain, loaded from the ACA, to generate an LDevID certificate based off device fields.
2. Sign the generated certificate using a private key (see `privateKey()` method in the ACA).

PR #814, opened 2024-07-22
[#709] Implementing LDevID generation

iadgovuser59:
This code implements generating an LDevID certificate (bound to a device's TPM) via the combined actions of both the provisioner and HIRS portal. Additionally, a new column is visible under the portal's Issued Certificates page, in addition to the generation option for LDevIDs visible under the policy page.

Closes several tickets: #709, #738, #739, and #740.


HIRS_Provisioner.NET/hirs/appsettings.json
```diff
@@ -2,6 +2,7 @@
   "auto_detect_tpm":  "TRUE",
   "aca_address_port": "https://127.0.0.1:8443",
   "efi_prefix": "",
+  "cert_prefix": "",
```
iadgovuser29, line 5: This cert_prefix option is confusing. efi_prefix is looking for the root of the EFI system partition. cert_prefix is looking for the exact directory to place files we generate. Consider renaming cert_prefix to better fit the value being requested. Example: certificate_output_directory

iadgovuser29 (changes requested):

iadgovuser26 (approved):

iadgovuser29 (approved):

Issue #864, opened 2024-10-29
Revert Checkstyle rule in Checkstyle.xml and Apply New Changes to HIRS_UTILS


ThatSilentCoder:
I recently modified one of the modules in checkstyles.xml to ignore checks for Java constants that begin with an underscore. However, after further research and discussions with my colleagues, I believe we should revert this rule and instead update the Java constants to align with recommended Java guidelines. This change will help maintain code consistency and adherence to best practices. Apply the new changes to the entire HIRS_UTILS module.

Revert the following ...
```
        <module name="ConstantName">
            <property name="format"
                      value="[A-Z_][A-Z0-9]*(_[A-Z0-9]+)*$"/>
        </module>
```

to this:

`<module name="ConstantName"/>`

PR #865, opened 2024-10-30
[#864] Undid Constant Name Rule In Checkstyles XML and Applied Formatting/Changes to HIRS_UTILS

ThatSilentCoder:
### Description
I recently modified one of the modules in checkstyles.xml to ignore checks for Java constants that begin with an underscore. However, after further research and discussions with my colleagues, I believe we should revert this rule and instead update the Java constants to align with recommended Java guidelines. This change will help maintain code consistency and adherence to best practices. Apply the new changes to the entire HIRS_UTILS module.

### Other Changes You Might Have Noticed:
Asides from undoing one of the checkstyle modules in the checkstyles.xml file, I also "lombok-ed" a lot of java files in the UTILS files and a good portion of the files in the HIRS_UTILS module were formatted according to checkstyles guidelines. 

### Test Instructions:
1. Run `./gradlew clean HIRS_UTILS:check --rerun-tasks` in your terminal window to verify there are no checkstyle errors.
2. Run `./gradlew clean build --rerun-tasks` in your terminal window to verify there are no build errors.
3. Go to the SwidTagsConstants.java file and verify none of the constants start with an underscore.

### DIFF
--- a/HIRS_AttestationCA/src/main/java/hirs/attestationca/persist/entity/userdefined/rim/BaseReferenceManifest.java
+++ b/HIRS_AttestationCA/src/main/java/hirs/attestationca/persist/entity/userdefined/rim/BaseReferenceManifest.java
@@ -103,7 +103,6 @@ public class BaseReferenceManifest extends ReferenceManifest {
      * @param rimBytes byte array representation of the RIM
      * @throws IOException if unable to unmarshal the string
      */
-    @SuppressWarnings("checkstyle:AvoidInlineConditionals")
     public BaseReferenceManifest(final String fileName, final byte[] rimBytes)
             throws UnmarshalException {
         super(rimBytes);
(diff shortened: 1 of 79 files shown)


chubtub (commented):
The gradle build succeeded but I got a lot of ant:checkstyle warnings, is that supposed to happen?

ThatSilentCoder:
> The gradle build succeeded but I got a lot of ant:checkstyle warnings, is that supposed to happen?

Yes, that is supposed to happen. When you run the check command on the HIRS_UTILS module, there should be no checkstyle warnings. But for the other modules, because none of the other PR's for checkstyle warnings have been merged into main as of yet, you are still going to get checkstyle warnings for all the other modules (asides from HIRS_UTILS) when you try to build the entire application.

chubtub (approved):

PR #868, opened 2024-11-05
Ensure hirs certs directory can be read by the mysql group

iadgovuser29:
This should let a new container read the exported certificates folder from another container.


iadgovuser29:
This didn't work.

Issue #848, opened 2024-10-21
Rename data structures to align with TCG documentation


chubtub:
The TCG documentation on firmware validation defines terminology to describe the related data structures and operations.  The ACA classes involved in firmware provisioning and validation have confusing method and variable names in this context.  These should be changed so that the code is more readable.

PR #886, opened 2024-12-10
[#879] Additional renaming and checkstyle  changes

### LINKED ISSUE #879: Rename ReferenceDigestValue
This is a follow-on to #848 and concerns renaming the ReferenceDigestValue class and table.

chubtub:
These changes follow those from #848.  It should be noted that after reviewing the documentation the ReferenceDigestValue class was deemed appropriately named and _not_  renamed as originally intended.  Variable names in a few other classes have been refactored to more clearly align with TCG documentation.  More checkstyle changes have been implemented as well.

Closes #879

### DIFF
--- a/HIRS_AttestationCA/src/main/java/hirs/attestationca/persist/validation/PcrValidator.java
+++ b/HIRS_AttestationCA/src/main/java/hirs/attestationca/persist/validation/PcrValidator.java
@@ -66,6 +66,13 @@ public class PcrValidator {
         baselinePcrs = Arrays.copyOf(pcrValues, TPMMeasurementRecord.MAX_PCR_ID + 1);
     }
 
+    /**
+     * Builds a string array of stored pcrs.
+     *
+     * @param pcrContent      string representation of the pcr content
+     * @param algorithmLength length of the algorithm
+     * @return string array representation of the stored pcrs.
+     */
     public static String[] buildStoredPcrs(final String pcrContent, final int algorithmLength) {
         // we have a full set of PCR values
         String[] pcrSet = pcrContent.split("\\n");
@@ -244,9 +251,9 @@ public class PcrValidator {
             // other information.
             String calculatedString = Hex.encodeHexString(
                     pcrInfoShort.getCalculatedDigest());
-            log.debug("Validating PCR information with the following:" +
-                    System.lineSeparator() + "calculatedString = " + calculatedString +
-                    System.lineSeparator() + "quoteString = " + quoteString);
+            log.debug("Validating PCR information with the following:"
+                    + System.lineSeparator() + "calculatedString = " + calculatedString
+                    + System.lineSeparator() + "quoteString = " + quoteString);
             validated = quoteString.contains(calculatedString);
             if (!validated) {
                 log.warn(calculatedString + " not found in " + quoteString);
--- a/HIRS_AttestationCA/src/main/java/hirs/attestationca/persist/validation/PcrValidator.java
+++ b/HIRS_AttestationCA/src/main/java/hirs/attestationca/persist/validation/PcrValidator.java
@@ -15,6 +15,7 @@ import org.apache.commons.codec.binary.Hex;
 import java.nio.charset.StandardCharsets;
 import java.security.NoSuchAlgorithmException;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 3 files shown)


HIRS_AttestationCAPortal/src/main/java/hirs/attestationca/portal/page/controllers/ReferenceManifestPageController.java
```diff
@@ -526,14 +522,18 @@ private void processTpmEvents(final List<SupportReferenceManifest> dbSupportRims
 
                             this.referenceDigestValueRepository.save(newRdv);
                         }
-                    } catch (CertificateException | NoSuchAlgorithmException | IOException e) {
+                    } catch (CertificateException e) {
```
ThatSilentCoder, line 525: Since we are printing the same stack trace for all three exceptions, why did you separate them out into their own catch statement? Do you believe that we will add more code to each exception case?

HIRS_AttestationCAPortal/src/main/java/hirs/attestationca/portal/page/controllers/ReferenceManifestDetailsPageController.java
```diff
@@ -608,6 +608,12 @@ public ModelAndView initPage(final ReferenceManifestDetailsPageParams params,
                 String uuidError = "Failed to parse ID from: " + params.getId();
                 messages.addError(uuidError);
                 log.error(uuidError, iaEx);
+            } catch (CertificateException cEx) {
```
ThatSilentCoder, line 611: I noticed that we are catching every possible exception the application could throw at us (with the very last catch statement on line 617 in the modified file), is that last catch necessary? If it is necessary, do you believe it would make sense to collapse all of the above exceptions into one big catch seeing as that we are logging the same exception and we aren't doing anything special per exception?

HIRS_AttestationCA/src/main/java/hirs/attestationca/persist/validation/PcrValidator.java
```diff
@@ -163,36 +163,36 @@ public StringBuilder validatePcrs(final String[] storedPcrs,
...
      */
     public List<TpmPcrEvent> validateTpmEvents(final TCGEventLog tcgMeasurementLog,
-                                               final Map<String, ReferenceDigestValue> eventValueMap,
+                                               final Map<String, ReferenceDigestValue> eventLogRecords,
                                                final PolicySettings policySettings) {
         List<TpmPcrEvent> tpmPcrEvents = new LinkedList<>();
         for (TpmPcrEvent tpe : tcgMeasurementLog.getEventList()) {
             if (policySettings.isIgnoreImaEnabled() && tpe.getPcrIndex() == IMA_PCR) {
-                log.info("IMA Ignored -> {}", tpe);
+                log.info(String.format("IMA Ignored -> %s", tpe));
```
ThatSilentCoder, line 176: I've noticed that you've changed the parameterized log messages to string format messages throughout the three files. Is this the standard that we are supposed to follow or is there another reason for this change?

ThatSilentCoder (changes requested):
I had some questions about some of the changes you"ve made. Asides from those questions, I didn't have any other issues with the PR.

HIRS_AttestationCAPortal/src/main/java/hirs/attestationca/portal/page/controllers/ReferenceManifestPageController.java
chubtub: This change was automatically applied by checkstyles.  There are instances in the tcg_rim_tool of exceptions broken out into distinct catch blocks for specific handling, which can be applied here if appropriate.

HIRS_AttestationCAPortal/src/main/java/hirs/attestationca/portal/page/controllers/ReferenceManifestDetailsPageController.java
chubtub: Exception handling can be more elegantly handled in the ACA.  In many cases a try block is followed by several catches with unique parameters that are otherwise identical.  If deemed appropriate these can be addressed in a future PR.

HIRS_AttestationCA/src/main/java/hirs/attestationca/persist/validation/PcrValidator.java
chubtub: This change was automatically applied by checkstyles so I assumed that was the standard we were following.  I can look into logging best practices to see if there's any reason to prefer one over the other.

ThatSilentCoder: No worries! I wasn't trying to be nit-picky. I wasn't entirely sure if this was the standard we wanted to stick to or if parameterized log messages was the way to go. I saw that the Intellij IDE suggested that as the "preferred" option; however I do not know what the team believes is best.

iadgovuser26 (approved):

iadgovuser26 (approved):

ThatSilentCoder (approved):
Overall looks good to me. Appreciate you answering my questions. Further discussions should be had on the subject of the log message formatting and the way we handle exceptions.

PR #893, opened 2025-01-14
[#849] Separate RIM and PCR value validations into distinct methods

### LINKED ISSUE #849: Separate RIM and FW validation
Currently, the FirmwareScvValidator class validates both the reference manifest (RIM) files and the firmware (FW) measurements at the same time.  This makes for a long method that is less readable; RIM and FW validation should be handled by separate methods.

chubtub:
These changes separate FirmwareScvValidator.validateFirmware() into two private methods, validateRimSignature() and validatePcrValues().

Closes #849


chubtub:
Pending #899

PR #993, opened 2025-08-25
Empty, null, Unknown, Not Specified equivalence in component serial and revision

iadgovuser29:
Unknown, empty, and null values for component serial and revision should be treated the same during component validation.

This PR also fixes some of the assumptions in the unit tests in SupplyChainCredentialValidatorTest.java. ComponentResultRepository was not being mocked. And multiple unit tests were using the incorrect component identifier object.


ThatSilentCoder (approved):
Had one question. Looks good to me!

PR #1012, opened 2025-09-25
removing access to vendor-table on file system, will only be accessib…

iadgovuser58:
…le from file in code


chubtub (approved):
Changes look good

PR #1051, opened 2025-11-03
[#1031] Migrate Issued Certs JSP Page To Thymeleaf

### LINKED ISSUE #1031: Migrate Issued Certificates JSP Page To Thymeleaf
The HTML structure for the **Issued Certificates** page has already been created. The page is under the `HIRS_AttestationCAPortal/src/main/resources/templates` directory.

The goal of this task is to complete the migration of this page from JSP to Thymeleaf. This includes wiring up the necessary data bindings, loops, conditionals, and any dynamic elements so that the page behaves and renders exactly as it did in the original JSP version.

Please ensure that all existing functionality, styling, and dynamic behaviors are preserved during the migration. This effort is part of a broader initiative to migrate all JSP-based views to Thymeleaf as the new templating engine.

Please also ensure that when you create a new pull request that you merge the changes into the **jsp-2-thymleaf-migration** branch.

iadgovuser58:
Converting issued-certificates page to Thymeleaf

Closes #1031

### DIFF
--- a/HIRS_AttestationCAPortal/src/main/resources/static/js/common.js
+++ b/HIRS_AttestationCAPortal/src/main/resources/static/js/common.js
@@ -137,9 +137,9 @@ function setDataTables(id, url, columns, options = {}) {
  *       certificate type.
  */
 function certificateDetailsLink(type, id, sameType) {
-  let href = "certificate-details?id=" + id + "&type=" + type;
-  let title = "";
+  const href = "certificate-details?id=" + id + "&type=" + type;
   let icon = iconPath;
+  let title = "";
 
   //If the details is the same certificate type use assignment icon,
   //otherwise use the icon for the certificate type.
@@ -166,7 +166,7 @@ function certificateDetailsLink(type, id, sameType) {
         break;
     }
   }
-  let html =
+  const html =
     "<a href=" +
     href +
     ">" +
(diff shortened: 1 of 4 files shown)


HIRS_AttestationCAPortal/src/main/resources/static/js/common.js
```diff
@@ -1,3 +1,5 @@
+let iconPath = "/icons";
```
ThatSilentCoder: If you don't intend on changing this variable in any part of the js file, declare it using the const keyword.

HIRS_AttestationCAPortal/src/main/resources/static/js/common.js
```diff
@@ -206,9 +207,9 @@ function rimDetailsLink(id) {
  * @param id of the certificate
  * @param pagePath path to the link
```
ThatSilentCoder: Make sure that the javascript docs parameters are in the same order as the parameters. Pretty much reverse the order of the two params.

HIRS_AttestationCAPortal/src/main/resources/templates/issued-certificates.html
```diff
@@ -1,6 +1,158 @@
 <!DOCTYPE html>
 <html lang="en" xmlns:th="http://www.thymeleaf.org" xmlns:layout="http://www.ultraq.net.nz/thymeleaf/layout"
     layout:decorate="~{layout}">
+
+<head>
+    <section layout:fragment="script">
+        <script th:src="@{/lib/jquery.spring-friendly/jquery.spring-friendly.js}"></script>
```
ThatSilentCoder: the jquery line shouldnt be needed since it's automatically imported when we decorate the page with the layout template

ThatSilentCoder: Ok, apologies. I assumed that it was already declared in the layout.html file. 

It's totally up to the team as to how we want to approach this because we can either declare that script line directly in the layout.html thereby saving us the trouble of adding this line in every single page we work on or we can do what we've done with all the pages so far and add it all the pages that need it (in this case: all the pages that use datatables).

PR #1054, opened 2025-11-06
[#1036] Converting Validation Reports page to Thymeleaf

### LINKED ISSUE #1036: Migrate Validation Reports JSP Page To Thymeleaf
The HTML structure for the **Validation Reports** page has already been created. The page is under the `HIRS_AttestationCAPortal/src/main/resources/templates` directory.

The goal of this task is to complete the migration of this page from JSP to Thymeleaf. This includes wiring up the necessary data bindings, loops, conditionals, and any dynamic elements so that the page behaves and renders exactly as it did in the original JSP version.

Please ensure that all existing functionality, styling, and dynamic behaviors are preserved during the migration. This effort is part of a broader initiative to migrate all JSP-based views to Thymeleaf as the new templating engine.

Please also ensure that when you create a new pull request that you merge the changes into the **jsp-2-thymleaf-migration** branch.

iadgovuser59:
This PR converts the Validation Reports page to Thymeleaf, alongside some other small changes to ensure API calls and helper functions work properly with the page. Closes #1036.

### DIFF
--- a/HIRS_AttestationCAPortal/src/main/resources/templates/fragments/download-info.html
+++ b/HIRS_AttestationCAPortal/src/main/resources/templates/fragments/download-info.html
@@ -2,7 +2,7 @@
 <html xmlns:th="http://www.thymeleaf.org">
 
 <body>
-    <th:block th:fragment="downloadIcon(id, label, customButtons, bodyContent)">
+    <th:block th:fragment="downloadInfo(id, label, customButtons, bodyContent)">
         <!-- Download icon anchor -->
         <a th:href="'#' + ${id}" data-toggle="modal" th:title="${label}">
             <img th:src="@{/icons/ic_file_download_black_24dp.png}" alt="Download Icon" />
--- a/HIRS_AttestationCAPortal/src/main/resources/templates/platform-credentials.html
+++ b/HIRS_AttestationCAPortal/src/main/resources/templates/platform-credentials.html
@@ -7,8 +7,8 @@
         <script th:src="@{/lib/jquery.spring-friendly/jquery.spring-friendly.js}"></script>
         <script th:inline="javascript">
            $(document).ready(function () {
-             let viewName = [[${page.viewName}]];
-             let url = viewName + "/list";
+             const viewName = [[${page.viewName}]];
+             const url = viewName + "/list";
 
              let columns = [
                {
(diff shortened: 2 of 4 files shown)


HIRS_AttestationCAPortal/src/main/resources/templates/validation-reports.html
```diff
@@ -1,292 +1,286 @@
...
+            return html;
+            }
+        </script>
+    </section>
+</head>
+
+<h1 layout:fragment="pageHeaderTitle">Validation Reports</h1>
+
+<div class="content" layout:fragment="content">
+
```
chubtub: Several other JSP pages have `<div class="aca-input-box-header">` here while the validation reports page did not.  Should it be here for consistency?

iadgovuser26 (changes requested):

HIRS_AttestationCAPortal/src/main/resources/templates/validation-reports.html
iadgovuser59: That is a good catch; I checked and the original JSP did not have it. I just tried changing this, but it appears that the header gets perhaps a bit large in terms of width (so I'm wondering if that's why it was originally like this). I figure that we could always return to it if necessary, but for now it seems that this should visually match the current contents of the old page.

HIRS_AttestationCAPortal/src/main/resources/templates/platform-credentials.html
```diff
@@ -7,7 +7,9 @@
         <script th:src="@{/lib/jquery.spring-friendly/jquery.spring-friendly.js}"></script>
         <script th:inline="javascript">
            $(document).ready(function () {
-             let url = "platform-credentials/list";
+             let viewName = [[${page.viewName}]];
```
ThatSilentCoder: Change URL and Viewname to const since they don't change at all throughout the function

HIRS_AttestationCAPortal/src/main/resources/templates/fragments/download-info.html
```diff
@@ -2,20 +2,20 @@
 <html xmlns:th="http://www.thymeleaf.org">
 
 <body>
-    <div th:fragment="downloadIcon(id, label, customButtons, bodyContent)">
+    <th:block th:fragment="downloadIcon(id, label, customButtons, bodyContent)">
```
ThatSilentCoder: change the name of DownloadIcon to downloadInfo or something else that makes sense to you.

iadgovuser58 (approved):

Issue #1018, opened 2025-09-29
Add remaining fields to EK Cert's TPM Security Assertions in ACA Portal


iadgovuser62:
This issue addresses the addition of remaining TPM Security Assertion fields on the ACA Portal's Certificate Details page for EK Certificates. For fields not provided, it will instead display something along the lines of "Not Present".

PR #1062, opened 2025-11-19
[#1018] Add remaining TPMSecurityAssertions fields to EK Cert + ExtendedKeyUsage values on ACA portal details page

iadgovuser62:
This PR addresses 2 additions on the Details pages for Endorsement Key Credentials uploaded to the ACA.

1. Updated the `Extended Key Usage` field to have a value that matches the respective OID. Previously, it just displayed the OID (i.e. `###.###.###...`). Now, a local Map has been implemented with expected OIDs that match a registry value, and that is displayed with the OID in a human-readable fashion (i.e. `###.###.### (tcg-random-spec)`). If new expected OIDs come up in the future they will be easy to add to this Map.

2. Added remaining TPM Security Assertions fields to the ACA portal details page for EK Certs, in accordance with [TCG EK Credential Profile for TPM Family 2.0](https://trustedcomputinggroup.org/resource/http-trustedcomputinggroup-org-wp-content-uploads-tcg-ek-credential-profile-v-2-5-r2_published-pdf/) (see page 24 of spec). Some notes on design decisions:
    - Previously, only the first 5 fields of the TPM Security Assertions were displayed. Now they all are, including Common Criteria and FIPS Level information. To conserve space, this section has nested collapsible subsections to view different Assertion fields.
    - Fields that are not included in an uploaded EK Cert will display `[Not Present]` as their value.
    - TPM Security Assertions includes many nested upon nested sequences. The most nested case is `Common Criteria` which contains `URIReference` sequences, which contains `AlgorithmIdentifier` sequences. Attempting to represent this structure perfectly in the current code would cause much complexity. To resolve this, all fields underneath `Common Criteria` have been flattened into that 1 class (`CommonCriteriaMeasures.java`) and no longer nested (i.e. `profileUri`, `profileAlgOid`, `targetUri`, etc.). Though this is less representative of the spec, it has allowed for much simpler parsing and retrieving. It also does NOT make use of the currently existing `URIReference.java`, as that class is in use by other parts of HIRS and may require complex refactoring if it was going to be used here. Complexity has been sacrificed for slight redundancy instead.
    - `EndorsementCrendential.java` is currently holding the newly displayed TPM Security Assertions fields as Transient variables, as opposed to adding them to `TPMSecurityAssertions.java` even though the latter would be a more accurate representation of the data. This has been done to avoid adding new columns to the backend database, which would alter the schema and require a database reset for other users. Being Transient variables, they do not persist in the database and are parsed for display only upon viewing the details page. This allows end users to update their ACA with these new fields without having to reset their databases, promoting backwards compatibility. In any case, since `EndorsementCredential` objects have only a single `TPMSecurityAssertions` field, adding the new fields to `EndorsementCredential` instead is functionally equivalent for our purposes. This same approach can be used for new fields added in the future.
    - Java ASN1 types have been changed to their equivalent basic Java types so that Hibernate/JPA can handle them. i.e. `CommonCriteriaMeasures.java` contains the `plus` field, which would ordinarily be `ASN1Boolean `type, but now has been changed to just `boolean` type. This was already being done in `TPMSecurityAssertions.java`.
    - For the StringBuilder methods and other parsers used to display information to ACA portal, HIRS consistently uses key-value Maps that are comma-separated. TPM Security Assertions contains 2 fields (`AlgorithmIdentifier Parameters` and `URIReference HashValue`) that are currently typed as Byte Arrays. Byte Arrays inherently contain commas and conflict with said comma-splitting functionality. Escaping those commas was looking to be very complex code for just a few fields. A corner-cutting decision was made to just replace all commas with spaces within those Byte Arrays. This resolved the conflict but also resulted in a display like the following example. Feedback would be appreciated if this display is or isn't appropriate.
       - Instead of this `[5, 6, 7, 8]`, the fields are displayed like this `[5 6 7 8]`.

Test Instructions:

1. Clone this PR's branch and use it to spin up a local ACA.
2. Upload a certificate chain and a respective Endorsement Key Credential cert that contains TPM Security Assertions fields.
3. On the details page of that cert, confirm that the "Extended Key Usage" field shows both the OID and the human-readable text value.
4. On the details page of that cert, expand the "TPM Security Assertions" field and confirm that the display matches the info included in the cert.

Resolves: #1018

### DIFF
--- a/HIRS_AttestationCA/src/main/java/hirs/attestationca/persist/entity/userdefined/certificate/EndorsementCredential.java
+++ b/HIRS_AttestationCA/src/main/java/hirs/attestationca/persist/entity/userdefined/certificate/EndorsementCredential.java
@@ -634,6 +634,7 @@ public class EndorsementCredential extends DeviceAssociatedCertificate {
      * a {@link CommonCriteriaMeasures} object with the discovered fields.
      *
      * @param ccSeq the ASN1Sequence containing Common Criteria information
+     * @throws IOException parsing individual subcomponents failed
      */
     private void parseCommonCriteria(final ASN1Sequence ccSeq) throws IOException {
         CommonCriteriaMeasures parsedCommonCriteria = new CommonCriteriaMeasures();
@@ -818,6 +819,7 @@ public class EndorsementCredential extends DeviceAssociatedCertificate {
      *
      * @param taggedObj the ASN1TaggedObject to manually parse
      * @return oidString the Oid to return in String form
+     * @throws IOException parsing individual subcomponents failed
      */
     private static String parseTaggedOid(final ASN1TaggedObject taggedObj) throws IOException {
         ASN1Primitive primitiveObj = unwrapTaggedObject(taggedObj);
--- a/HIRS_AttestationCAPortal/src/main/resources/templates/certificate-details.html
+++ b/HIRS_AttestationCAPortal/src/main/resources/templates/certificate-details.html
@@ -582,15 +582,11 @@
                     <div class="certrow">
                         <div class="col-md-1 col-md-offset-1">
                             <span class="colHeader">
-                                <a role="button" data-bs-toggle="collapse" class="collapsed" href="#tpmSpecificationInner"
-                                   aria-expanded="true" data-bs-placement="top" aria-controls="tpmSpecificationInner">
-                                    TPM Specification ▼
-                                </a>
+                                TPM Specification
                             </span>
                         </div>
                         <div id="tpmSpecification" class="col col-md-8">
-                            <div id="tpmSpecificationInner" class="accordion-body card-body collapse" role="tabpanel"
-                                 aria-expanded="false">
+                            <div id="tpmSpecificationInner" class="accordion-body card-body">
                                 <div>Family:&nbsp;<span>[[${initialData.get('TPMSpecificationFamily')}]]</span></div>
                                 <div>Level:&nbsp;<span>[[${initialData.get('TPMSpecificationLevel')}]]</span></div>
                                 <div>Revision:&nbsp;<span>[[${initialData.get('TPMSpecificationRevision')}]]</span></div>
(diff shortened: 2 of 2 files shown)


HIRS_AttestationCA/src/main/java/hirs/attestationca/persist/entity/userdefined/certificate/EndorsementCredential.java
```diff
@@ -351,39 +399,51 @@ private void parseSequence(final ASN1Sequence seq, final boolean addToMapping,
...
-                                .getValue().intValue();
-                        if (ekCertGenLocVal >= EK_LOC_VAL_MIN
-                                && ekCertGenLocVal <= EK_LOC_VAL_MAX) {
-                            TPMSecurityAssertions.EkGenerationLocation ekCertGenLoc
-                                    = TPMSecurityAssertions.EkGenerationLocation.
-                                    values()[ekCertGenLocVal];
-                            tpmSecurityAssertions.setEkCertificateGenerationLocation(ekCertGenLoc);
-                        }
+                    switch (tag) {
+                        case EK_TYPE_TAG -> {
```
ThatSilentCoder: I like the switch statements instead of the if/else conditions

HIRS_AttestationCA/src/main/java/hirs/attestationca/persist/entity/userdefined/certificate/EndorsementCredential.java
```diff
@@ -568,4 +628,213 @@ private void parseSingle(final ASN1Primitive component, final boolean addToMappi
...
+        return octetString.toASN1Primitive();
+    }
+
+    /**
+     * Parses an ASN.1 tagged object to extract an Object Identifier (OID) as a string.
+     *
+     * @param taggedObj the ASN1TaggedObject to manually parse
+     * @return oidString the Oid to return in String form
+     */
+    private static String parseTaggedOid(final ASN1TaggedObject taggedObj) throws IOException {
```
ThatSilentCoder: Make sure to add @throws to your JavaDoc

HIRS_AttestationCAPortal/src/main/resources/templates/certificate-details.html
```diff
@@ -603,23 +603,252 @@ <h2 id="headingTwo" class="accordion-header">
...
+                                                )"></span>
+                                            </td>
+                                        </tr>
+                                    </table>
                                 </div>
-                                <div>ek Generation
-                                    Location:&nbsp;<span>[[${initialData.get('TPMSecurityAssertionsEkGenerationLocation')}]]</span></div>
-                                <div>ek Certificate Generation
-                                    Location:&nbsp;<span>[[${initialData.get('TPMSecurityAssertionsEkCertificateGenerationLocation')}]]</span></div>
+                                <table class="table table-bordered table-sm">
```
ThatSilentCoder: Do you think this table in this section should also be hidden until someone clicks on the "expand section button" or do you think it should stay the same?

ThatSilentCoder (changes requested):
I uploaded all the certificates you gave me and some that I had on my system and it worked well. I did have some questions about certain parts of the code.

HIRS_AttestationCAPortal/src/main/resources/templates/certificate-details.html
```diff
@@ -603,23 +603,252 @@ <h2 id="headingTwo" class="accordion-header">
...
                                 </a>
                             </span>
                         </div>
                         <div id="tpmSecurityAssertion" class="col col-md-8">
                             <div id="tpmSecurityAssertionInner" class="accordion-body card-body collapse" role="tabpanel"
                                  aria-expanded="false">
-                                <div>Version:&nbsp;<span>[[${initialData.get('TPMSecurityAssertionsTpmSecAssertsVersion')}]]</span></div>
-                                <div>Field
-                                    Upgradeable:&nbsp;<span>[[${initialData.get('TPMSecurityAssertionsFieldUpgradeable')}]]</span>
+                                <table class="table table-bordered table-sm">
```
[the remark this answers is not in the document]
iadgovuser62: Same response as for the one above.

PR #1100, opened 2026-02-03
[#1099] Update Application To Support JDK 25

### LINKED ISSUE #1099: Migrate to JDK 25 and Gradle 9.1.0
In order to support newer signature schemes (such as ML-DSA), we will need to upgrade the HIRS JDK version to a newer version, which is JDK 25 in this case. 

This will also require a migration to Gradle 9.1.0 (minimum supported version).

_Note that this will likely be a fairly involved issue, as it will encompass package updates, as well as deprecated Java features._

ThatSilentCoder:
### Description
Make changes to Gradle and other configuration files to support the transition to Java 25.

### Test Instructions:
1. Before following any of the subsequent steps, you will need to install the latest JDK 25 on your system. 

If you are on a RHEL-based Linux distro:

```
sudo dnf install java-25-openjdk
sudo dnf install java-25-openjdk-devel
```

If you are on DEBIAN-based Linux distro:
```
sudo apt install openjdk-25-jdk
sudo apt install java-25-openjdk-devel
```

2. Change the Java and Javac versions to 25:

```
sudo alternatives --config java

sudo alternatives --config javac
```

3. Confirm that you have Java and Javac 25 installed on your system of choice
```
java --version
javac --version
```

4. Run the ACA using your preferred method and confirm that the application can run on your system of choice.
5. Confirm that you can upload files, delete items off pages, interact with the website and store items in the ACA.
6. Confirm that no errors related to JDK25 appear either in the logs or on the console.
7. Confirm that when you execute `./gradlew build`, all the tests pass and there are no compliation errors.

### Issues This PR Addresses
Closes #1099


build.gradle
```diff
@@ -5,9 +5,10 @@ import java.util.concurrent.TimeUnit
 plugins {
     // Apply the application plugin to add support for building a CLI application in Java.
     id 'application'
-    id 'com.github.spotbugs' version '6.0.13' apply false
+    id 'com.github.spotbugs' version '6.4.8' apply false
     id 'org.owasp.dependencycheck' version '12.2.0'
     id 'io.spring.dependency-management' version '1.1.7' apply false
```
iadgovuser59: Not directly related to this PR, but I think we can remove this plugin (Spring Dependency Management). Per their docs, I don't think that we're using any patterns involving this; I believe that this might even add slightly extra build time, so removing could remove overhead.

build.gradle
```diff
@@ -5,9 +5,10 @@ import java.util.concurrent.TimeUnit
...
+    id 'com.palantir.git-version' version '4.3.0'
```
iadgovuser59: I think that we might perhaps want to examine if we need to use a plugin for this or not; initially, this one did use an included JGit library, but newer versions supposedly require `git` to be on the path.

Functionally speaking, this is not much different than the old `git exec` pattern (both require a real `git` executable to be on the path/machine), so I'm not sure if we need to add an extra library just for the Git hash.

build.gradle
```diff
@@ -101,29 +103,21 @@ subprojects {
...
+project.ext["projVersion"] = projectVersion
 
-def buildTime = { ->
-    Date latestdate = new Date()
-    def time = latestdate.getTime()
-    long seconds = TimeUnit.MILLISECONDS.toSeconds(time)
-    return seconds
-}
-
-def gitHash = { ->
```
iadgovuser59: See above comment (it might be easier to reinstate this code).

iadgovuser59 (changes requested):
Overall this PR looks good. I submitted a number of small changes, though the core upgrade works for me.

build.gradle
ThatSilentCoder: I attempted to retrieve it using the non-plugin method, but for some reason, it turned out to be more troublesome than it was worth. Unfortunately, after the removal of the exec keyword in Gradle, I was forced to adopt their approach. This ended up splitting the process into two tasks, and it didn’t work as expected. From what I understand, this particular plugin is lightweight and adds only a handful of useful Git-related tasks

build.gradle
ThatSilentCoder: Removed plugin and it still works. Good call!

iadgovuser59 (approved):

PR #1127, opened 2026-03-06
[#1090] Fix ACA Provisioning Issue When Using ECC Key

### LINKED ISSUE #1090: Fix ACA provision using ecc key
When using aca_setup.sh using the -aa ecc parameter it sets the ACA certificate chain to and ecc 512 with sha384. There is an issue when provisioning using an ecc cert:
~~~
ERROR : Servlet.service() for servlet [dispatcherServlet] in context with path [] threw exception [Request processing failed: hirs.attestationca.persist.exceptions.CertificateProcessingException: Encountered error while generating identity credential: cannot create signer: Supplied key (sun.security.ec.ECPrivateKeyImpl) is not a RSAPrivateKey instance] with root cause
java.security.InvalidKeyException: Supplied key (sun.security.ec.ECPrivateKeyImpl) is not a RSAPrivateKey instance
~~~ 
The ACA needs to be updated to handle ecc keys, or more importantly made crypt agile so it can use mldsa keys in the future.

Note that the ACA will currently show the provision as passing, yet won't create a new cert for the client. ACA should show that as a failure.

ThatSilentCoder:
### Description
Fix exceptions caused by provisioning against the ACA that is setup using the ECC key parameter in the setup script.

---

### Test Instructions:
1. Remove your local setup of the ACA using the follwing commands:

On Linux: 
```
sudo package/linux/aca/aca_remove_setup.sh
```

On Windows (Powershell 7 Core in Admin mode):
```
package\win\aca\aca_remove_setup.ps1
```

2. Setup the ACA to use ECC keys as follows:

On Linux: 
```
sudo package/linux/aca/aca_setup.sh -aa ecc -u
```

On Windows (Powershell 7 Core in Admin mode):
```
package\win\aca\aca_setup.ps1 -aa ecc -u
```

3. If everything is configured correctly, run the ACA.
4. Run the provisioner against the ACA for different policy settings: no policy enabled, platform credential
validation enabled, endorsement credential validation enabled, firmware validation enabled.
5. Verify that for each policy setting, assuming you've uploaded the correct certificates/RIMS up to the ACA per policy, the provisioning succeeds and that the ACA produces two issued certificates.
6. Verify that for each policy setting, no exceptions/errors are thrown in the console/logs.

---
### Summary Of Changes:

- Added the `-aca-alg`, `-ta-alg`, and `-db-alg` parameters to the PowerShell scripts to enable ACA setup on Windows using both ECC and RSA. Since introducing these parameters required updates across multiple scripts, additional fixes were made to address improper PowerShell syntax, along with general formatting improvements to the .ps1 files.

- Used the ShellCheck plugin while updating portions of the Bash scripts to address warnings and potential issues. Also improved the help menu output for better usability.

- ACA can now determine the type of keys being provided, rather than assuming all keys are RSA, and supports handling both ECC and RSA during the provisioning process.
---

### Issues This PR Addresses:
Closes #1090

### DIFF
--- a/HIRS_AttestationCA/src/main/java/hirs/attestationca/persist/entity/userdefined/certificate/IDevIDCertificate.java
+++ b/HIRS_AttestationCA/src/main/java/hirs/attestationca/persist/entity/userdefined/certificate/IDevIDCertificate.java
@@ -1,9 +1,12 @@
 package hirs.attestationca.persist.entity.userdefined.certificate;
 
 import hirs.attestationca.persist.entity.userdefined.Certificate;
+import hirs.attestationca.persist.entity.userdefined.certificate.attributes.DiceCertificateInfo;
+import hirs.attestationca.persist.entity.userdefined.certificate.attributes.DiceCertificateParser;
 import jakarta.persistence.Column;
 import jakarta.persistence.Entity;
 import jakarta.persistence.Transient;
+import jakarta.persistence.PostLoad;
 import lombok.EqualsAndHashCode;
 import lombok.Getter;
 import lombok.extern.log4j.Log4j2;
@@ -20,6 +23,7 @@ import org.bouncycastle.cert.jcajce.JcaX509ExtensionUtils;
 
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
+import java.io.Serial;
 import java.nio.file.Path;
 import java.time.Instant;
 import java.util.HashMap;
@@ -55,8 +59,16 @@ public class IDevIDCertificate extends Certificate {
     private static final String POLICY_QUALIFIER_VERIFIED_TPM_FIXED = "2.23.133.11.1.2";
     private static final String POLICY_QUALIFIER_VERIFIED_TPM_RESTRICTED = "2.23.133.11.1.3";
 
+    /**
+     * The raw byte array of the subject alternative name extension, if present.
+     * This will be null if the certificate does not contain a subject alternative name extension.
+     */
     @Transient
     private byte[] subjectAltName;
+
+    /** Parsed DICE attributes from the certificate, if present. */
+    @Transient
+    private transient DiceCertificateInfo diceCertificateInfo;
 
     /**
      * Corresponds to the hwType field found in a Hardware Module Name (if present).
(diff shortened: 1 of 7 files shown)


HIRS_AttestationCA/src/main/java/hirs/attestationca/persist/provision/service/AttestationCertificateAuthorityService.java
```diff
@@ -1,9 +1,9 @@
-package hirs.attestationca.persist;
+package hirs.attestationca.persist.provision.service;
 
 /**
- * Defines the responsibilities of the Attestation Certificate Authority.
+ * Interface that defines the responsibilities of the Attestation Certificate Authority service.
  */
-public interface RestfulInterface {
+public interface AttestationCertificateAuthorityService {
```
ThatSilentCoder: Renamed the interface class to better reflect its purpose.

HIRS_AttestationCA/src/main/java/hirs/attestationca/persist/provision/service/CertificateRequestProcessorService.java
```diff
@@ -387,4 +448,88 @@ private AppraisalStatus.Status doQuoteValidation(final Device device) {
...
+                    + "identity credential: " + exception.getMessage(), exception);
+        }
+    }
+
+    /**
+     * Returns a string representation of the signature algorithm that will be used with the provided private key.
+     *
+     * @return signature algorithm associated with the private key
+     */
+    private String getSignatureAlgorithm() {
```
ThatSilentCoder: This function resolves the issue.

iadgovuser59 (changes requested):
I would say that the core shell scripts and TLS portion look good as written, but I'm wondering if we might need to clarify the scope of this PR a bit?

I'd say that the ProvisionUtils.java ECC changes appear to be more of an experimental piece which may require coordination with the Provisioner (currently with some RSA-shaped logic), so my suggestion might be to trim down the PR to only the required portions (perhaps removing or separating some of the refactoring and TPM-specific ECC provisioning portions for now).

PR #1179, opened 2026-04-20
Update exception handling in TCG Event Log when event log can't parse

iadgovuser58:
Closes 1168.

Event logs that cannot be parsed produce exceptions that don't get properly handled.

This issue covers:

1. Divides the exceptions in TCGEventLog into "digest parsing errors" (-> validation cannot continue) and "event content parsing errors" (-> validation can continue). 
2. TCG Event Log will now only send up either an IO exception or a Runtime exception , either of which will require that validation stops. Any other exceptions will be caught and handled in the TCG Event Log or below.
- In event header fields: More robust error checking is done on the event header fields so that if the file is corrupt, this will be caught before an exception is thrown (such as a length field that is larger than the remaining bytes of the file). That error will be noted and thrown as an IO exception from the TCGEventLog to the caller, to stop attestation.
- In event content: This is now surrounded by a try/catch, and any error that occurs here (other than a runtime error) will be caught and logged but not sent from the TCGEventLog to the caller.
3. Additionally, clarify code, variable names, etc. when it comes to event header vs content.

This issue does not cover:
- A future issue will need to use the exception updates presented in this issue to base the decision to not upload/process event log file in ACA & display.

To test:
- Using the eventlog tool print function, test as many RIMs as possible, to include RIMs with errors. 
(Note for testing: This issue does not yet fix the ACA RIM upload error.)

### DIFF
--- a/HIRS_AttestationCA/src/main/java/hirs/attestationca/persist/entity/userdefined/rim/EventLogMeasurements.java
+++ b/HIRS_AttestationCA/src/main/java/hirs/attestationca/persist/entity/userdefined/rim/EventLogMeasurements.java
@@ -14,8 +14,6 @@ import lombok.Setter;
 import lombok.extern.log4j.Log4j2;
 
 import java.io.IOException;
-import java.security.NoSuchAlgorithmException;
-import java.security.cert.CertificateException;
 import java.util.ArrayList;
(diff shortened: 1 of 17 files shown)


chubtub (approved):
Changes look good

PR #1182, opened 2026-04-24
Update exception handling & info to aca/display when event log can't parse

### LINKED ISSUE #1178: Update exception handling & info to aca/display when event log can't parse
Previous issues divided the exceptions from TCGEventLog into "digest errors" (validation cannot continue) and "event content errors" (validation continues), and only passes on exceptions for "digest errors". This issue captures that exception in the ACA/GUI so that an event log that fails to parse the critical items (ie. digests) does not get uploaded and a message is displayed to the user.

iadgovuser58:
Closes #1178.

Event logs that cannot be parsed produce exceptions that don't get properly handled. 

Previous 2 related PRs divided the exceptions from TCGEventLog into "digest errors" (validation cannot continue) and "event content errors" (validation continues), and only passes on exceptions for "digest errors". 

This PR:

1. Previous work during thymeleaf migration pulled some of the parsing labor from the controllers into service classes. This issue pulled additional parsing labor from the reference manifest controller into the reference manifest service class. 
2. During the initial check of the support rim, this issue attempts to parse it as a TCGEventLog; if any exceptions are thrown, this means that the event log failed to parse the critical items (ie. digests). This could be due to a bad input event log, or due to coding errors in parsing. Either way, it stops the upload and prints a popup error message to the ACA webpage, but also does not crash the ACA. 
3. Future issue will need to address checking the support RIM to differentiate TCGEventLog from any other type of support RIM.

To test:
Check many event log uploads to the ACA. Check bad event logs as well - check some that simply have bad event content (those should still upload) and check some that cannot parse the digests (these should not upload, but should not crash the ACA).

### DIFF
--- a/HIRS_AttestationCA/src/main/java/hirs/attestationca/persist/dto/PageMessages.java
+++ b/HIRS_AttestationCA/src/main/java/hirs/attestationca/persist/dto/PageMessages.java
@@ -1,4 +1,4 @@
-package hirs.attestationca.portal.page;
+package hirs.attestationca.persist.dto;
 
 import lombok.Getter;
 
--- a/HIRS_AttestationCA/src/main/java/hirs/attestationca/persist/dto/package-info.java
+++ b/HIRS_AttestationCA/src/main/java/hirs/attestationca/persist/dto/package-info.java
@@ -1,0 +1,1 @@
+package hirs.attestationca.persist.dto;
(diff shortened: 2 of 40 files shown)


chubtub (commented):
Changes look good as they currently stand.  Approval is pending further testing of the ACA.

chubtub (dismissed):
Changes look good pending further testing of the ACA.

chubtub (approved):
Changes look good

PR #1189, opened 2026-05-07
V3.2 update ci rims

chubtub:
For each base RIM test pattern under .ci/, create a JSON config file in the directory above, and then regenerate the base RIM to remove all whitespacing.


chubtub:
This PR was meant to separate the updating of test patterns from code changes.  These are too intertwined to be practically separated, please see #1172 for all related changes.

PR #1208, opened 2026-06-02
Modify regex to allow uploads of RIM filenames with spaces

iadgovuser62:
Previously would not allow uploads of Base RIMs or Support RIMs that have spaces in filename.


chubtub (approved):
Changes look good.

iadgovuser26 (approved):

PR #1210, opened 2026-06-04
[#1207] Create links to base RIMs in composite RIM payload

chubtub:
### Major changes ###

1. Composite RIMs now link to lower base RIMs in their payloads.
2. Payload support RIMs display expected PCRs or an error message; base RIMs do not.
3. Payload files that cannot be queried in the backend are logged for reference.

Closes #1207

PR #1215, opened 2026-06-11
Prevent non-unique returns for RIM lookup when reading composite SWID resources

iadgovuser62:
RIM Details Page Service has functionality to read through SWID file's resources, which typically contain other RIMs of different types. Originally, it queried the database to look up those RIMs by their hashes AND rimType. #1210 modified this query to just look up by the hash initially. This change caused the possibility of non-unique returns, which in turn caused an error in the ACA.

This PR reverts the query so it looks up RIMs by both their hash AND rimType, while still retaining functionality added by #1210 .


chubtub (approved):
Changes look good

PR #1231, opened 2026-06-24
[#1228] Label composite RIMs on the ReferenceManifest page

### LINKED ISSUE #1228: RIM page should identity composite RIMs
Image

This table should mark the top-level base RIM.

chubtub:
The major change here is that base RIMs with Rim Type "Base - Hybrid" are now Rim Type "Composite".

image

Closes #1228


iadgovuser26:
It was decided to leave the type as Base-hybrid to match the specification,

PR #1249, opened 2026-08-06
[#1239] Implement hashing algorithm for support RIM

### LINKED ISSUE #1239: Support hash algorithms for pcrims other than SHA256
Currently, SHA256 is hard-coded into the RIM create function.  SHA384 and 512 should be supported as options.

chubtub:
Major changes include:

- Add constants to specify SHA256, SHA384, and SHA512
- Add methods to calculate a file digest using the above hashes
- Create a base RIM with a file attribute corresponding to the above hash and digest

Closes #1239

### DIFF
--- a/HIRS_Utils/src/main/java/hirs/utils/swid/SwidTagGateway.java
+++ b/HIRS_Utils/src/main/java/hirs/utils/swid/SwidTagGateway.java
@@ -528,27 +528,28 @@ public class SwidTagGateway {
         String fileHash = jsonObject.getString(SwidTagConstants.HASH, "");
 
         if (fileHash.isEmpty()) {
-            if (rimEventLog.isEmpty() || hashAlgorithm.isEmpty()) {
+            if (rimEventLog.isEmpty()) {
                 throw new RuntimeException("The config file must declare a file hash, "
-                       + "or a support RIM and hash algorithm must be given.");
-            } else {
-                switch (hashAlgorithm.toUpperCase()) {
-                    case SwidTagConstants.SHA256:
-                        fileHash = HashSwid.get256Hash(rimEventLog);
-                        hashKey = SwidTagConstants.SHA_256_HASH;
-                        break;
-                    case SwidTagConstants.SHA384:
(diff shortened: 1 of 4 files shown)


HIRS_Utils/src/main/java/hirs/utils/swid/SwidTagGateway.java
```diff
@@ -511,20 +518,39 @@ private Directory createDirectory(final JsonObject jsonObject) {
...
         file.setName(jsonObject.getString(SwidTagConstants.NAME, ""));
         file.setSize(new BigInteger(jsonObject.getString(SwidTagConstants.SIZE, "0")));
         Map<QName, String> attributes = file.getOtherAttributes();
-        String fileHash;
-        if (rimEventLog.isEmpty()) {
-            fileHash = jsonObject.getString(SwidTagConstants.SHA_256_HASH.getLocalPart());
-        } else {
-            fileHash = jsonObject.getString(SwidTagConstants.SHA_256_HASH.getLocalPart(),
-                    HashSwid.get256Hash(rimEventLog));
+        QName hashKey = new QName("");
```
iadgovuser59: I'm currently seeing that this is causing an error in pcrim_tests.sh in RIM-Tool [#56](https://github.com/nsacyber/RIM-Tool/pull/56), presumably due to the lack of a default value if `--algorithm` is omitted. If I change this to SwidTagConstants.SHA_256_HASH, this works properly.

iadgovuser59 (commented):
I think this looks good to me except for the error I'm seeing in pcrim_tests.sh in RIM-Tool [#56](https://github.com/nsacyber/RIM-Tool/pull/56).

I'm also curious if we should look into perhaps supporting a field for the algorithm in the Base RIM JSON config (eg `"algorithm"`) since at present, `"hash"` in the JSON has no way to specify an algorithm or use this with the `--algorithm` parameter. If we don't allow the param to be specified, we can't necessarily infer the hash type just from the Base64 string; ie this would always default to SHA-256.

iadgovuser59 (approved):

